PR #40, opened 2019-12-05
feat: add label fullStacktrace

### LINKED ISSUE #34: Stacktrace should not be written to log message
Hello.

I've noticed pretty strange things. Somehow you are adding whole stack trace information to textPayload which is weird. Also  it look very ugly in stackdriver logging web UI because you just put there extremely long string that can not be displayed in nice and verbose way.

I think that nobody wants to see it in textPayload. Why you can't add full stacktrace to property/label named "stacktrace" ? 

As for textPayload the only one reasonable modification is to add only message from exception at new line

This is part of the code that does all this things

```
  private LogEntry logEntryFor(ILoggingEvent e) {
    StringBuilder payload = new StringBuilder(e.getFormattedMessage()).append('\n');
    writeStack(e.getThrowableProxy(), "", payload);

    Level level = e.getLevel();
    LogEntry.Builder builder =
        LogEntry.newBuilder(Payload.StringPayload.of(payload.toString().trim()))
            .setTimestamp(e.getTimeStamp())
.setSeverity(severityFor(level));
```

athakor:
Fixes #34

### DIFF
--- a/google-cloud-contrib/google-cloud-logging-logback/src/test/java/com/google/cloud/logging/logback/LoggingAppenderTest.java
+++ b/google-cloud-contrib/google-cloud-logging-logback/src/test/java/com/google/cloud/logging/logback/LoggingAppenderTest.java
@@ -113,10 +113,10 @@ public class LoggingAppenderTest {
             .setSeverity(Severity.ERROR)
             .setLabels(
                 new ImmutableMap.Builder<String, String>()
-                    .put("levelName", "ERROR")
-                    .put("levelValue", String.valueOf(40000L))
-                    .put("loggerName", this.getClass().getName())
-                     .put("fullStacktrace", "this is a test".trim())
+                    .put("levelName","ERROR")
+                    .put("levelValue",String.valueOf(40000L))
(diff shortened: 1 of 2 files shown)


google-cloud-contrib/google-cloud-logging-logback/src/test/java/com/google/cloud/logging/logback/LoggingAppenderTest.java
```diff
@@ -115,6 +116,7 @@ public void testFilterLogsOnlyLogsAtOrAboveLogLevel() {
                     .put("levelName", "ERROR")
                     .put("levelValue", String.valueOf(40000L))
                     .put("loggerName", this.getClass().getName())
+                     .put("fullStacktrace", "this is a test".trim())
```
chingor13: nit: there's an extra space

athakor: done

Issue #16, opened 2019-11-19
Flaky integration tests


chingor13:
Example: https://source.cloud.google.com/results/invocations/ac171783-2001-49a9-94da-9f249f03c961/targets/github%2Fjava-logging%2Fgoogle-cloud-logging%2Ftarget%2Ffailsafe-reports/tests

pmakani:
@codyoss @chingor13 I think it's still flaky or could be problem in backend because of below code sometimes works and sometimes not. Is it still reproducible for you ?

```JAVA
public class LoggingExample {
  private Logging logging;
  private String logId;

  public LoggingExample() {
    this.logging = LoggingOptions.getDefaultInstance().getService();
    this.logId = "test-log-" + UUID.randomUUID().toString().substring(0, 8);
  }

  public void createLogEntry() {
    LogEntry firstEntry =
        LogEntry.newBuilder(Payload.StringPayload.of("test integration flakyness"))
            .setLogName(logId)
            .setResource(
                MonitoredResource.newBuilder("global")
                    .addLabel("project_id", logging.getOptions().getProjectId())
                    .build())
            .build();
    logging.write(Collections.singleton(firstEntry));
    logging.flush();
    System.out.println("createLogEntry logId :" + logId);
  }

  public void listLogEntries() {
    Page<LogEntry> entries =
        logging.listLogEntries(
            Logging.EntryListOption.filter(
                "logName=projects/" + logging.getOptions().getProjectId() + "/logs/" + logId));
    Iterator<LogEntry> entryIterator = entries.iterateAll().iterator();
    System.out.println("-----------------------");
    System.out.println("fetching listLogEntries");
    while (entryIterator.hasNext()) {
      System.out.println(entryIterator.next());
    }
  }

  public void deleteLogEntry() throws InterruptedException {
    Thread.sleep(5000);
    boolean deleted = logging.deleteLog(logId);
    System.out.println("--------------------------");
    System.out.println("is it deleted ? " + deleted);
  }

  public static void main(String[] args) throws InterruptedException {
    LoggingExample log = new LoggingExample();
    log.createLogEntry();
    log.listLogEntries();
    log.deleteLogEntry();
  }
}
```

PR #53, opened 2019-12-26
chore(deps): group core dependencies

chingor13:


chingor13:
Integration failure is related to #16

Issue #71, opened 2020-01-14
Support Cloud Run Monitored Resource


meltsufin:
**Is your feature request related to a problem? Please describe.**

Spring Cloud GCP Logging does not work with Cloud Run. The logs end up in the "Global" channel rather than the one for Cloud Run.

**Describe the solution you'd like**

The Cloud Run resource type `cloud_run_revision` should be automatically assigned to the log entries, similarly to how it's done for App Engine. See: [MonitoredResourceUtil](https://github.com/googleapis/java-logging/blob/master/google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java).

**Describe alternatives you've considered**

Using JSON-based logging on Cloud Run. It works better but has other issues like not supporting `TRACE` and `WARN` levels.

**Additional context**

Repro: https://github.com/meltsufin/cloud-run-helloworld

saturnism:
Ideally, there is a single public API call to retrieve/produce a MonitoredResource.  This is a common library as oppose to only for logs.

PR #78, opened 2020-01-22
fix: support for Cloud Run monitored resource

meltsufin:
Fixes: #71.

### DIFF
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
@@ -185,10 +185,10 @@ public class MonitoredResourceUtil {
 
   /* Detect monitored Resource type using environment variables, else return global as default. */
   private static Resource getAutoDetectedResourceType() {
-    if (System.getenv("K_SERVICE") != null &&
-        System.getenv("K_REVISION") != null &&
-        System.getenv("K_CONFIGURATION") != null &&
-        System.getenv("KUBERNETES_SERVICE_HOST") == null) {
+    if (System.getenv("K_SERVICE") != null
+        && System.getenv("K_REVISION") != null
+        && System.getenv("K_CONFIGURATION") != null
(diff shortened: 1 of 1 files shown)


google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -40,10 +40,13 @@
     ContainerName("container_name"),
     InstanceId("instance_id"),
     InstanceName("instance_name"),
+    Location("location"),
```
averikitsch: Is there a reason to use "location" instead of "region"?

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -171,6 +185,11 @@ private static String getValue(Label label) {
 
   /* Detect monitored Resource type using environment variables, else return global as default. */
   private static Resource getAutoDetectedResourceType() {
+    if (System.getenv("K_SERVICE") != null &&
```
averikitsch: Steren's note: Note that checking the presence of K_SERVICE and other env vars is not sufficient to differentiate between Cloud Run and any other Knative installation (e.g. Cloud Run for Anthos).

Would adding a check for "KUBERNETES_SERVICE_HOST" == null, fix this issue?

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
meltsufin: I looked at log entries that get ingested from stdout, and they all have the `location` attribute. So, I tried to match that.

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
meltsufin: Yes, that makes sense.

averikitsch:
@chingor13 can you approve?

PR #124, opened 2020-03-30
chore: Migrate logging synth.py to bazel

vam-google:
This PR migrates only synth.py but does not commit the regenerated files. The generation was tested and it works, the updated files are not commited due to breaking changes not related to bazel migration. 

**There is already an open PR with those braking changes: https://github.com/googleapis/java-logging/pull/111**


vam-google:
@chingor13 PTAL

PR #225, opened 2020-09-01
test: split tests by resource type. fix query time by limiting by timestamp

chingor13:
This should reduce the time for the integration tests from ~20 minutes to ~5 minutes and should prevent the tests from timing out.

* This PR breaks the monolilthic IT test file into tests by scope (java.util.Logger implementation, metrics, sinks, and direct logging client use).
* Previously, all the tests were defined in the BaseSystemTest with a single implementation which didn't seem necessary. Now the base class has shared helper functionality for fetching logs and instantiating the logging client.
* When fetching logs, we now only look at timestamps > 1 hour ago. This speeds up the list queries from 2 minutes to seconds.


chingor13:
ListSink tests fixed in #226

stephaniewang526 (approved):

PR #227, opened 2020-09-02
samples: limit list log in samples

chingor13:
Updates the `logging_list_log_entries` sample to limit logs fetched by timestamp. Querying over the entire log history can take minutes.


samples/snippets/src/main/java/com/example/logging/ListLogs.java
```diff
@@ -35,7 +35,8 @@ public static void main(String... args) throws Exception {
 
     try (Logging logging = options.getService()) {
 
-      String logFilter = "logName=projects/" + options.getProjectId() + "/logs/" + logName;
+      String logFilter = "logName=projects/" + options.getProjectId() + "/logs/" + logName +
+          " AND timestamp>=\"2020-09-01T00:00:00.000Z\"";
```
chingor13, line 39: Note that the code to generate this timestamp string is a bit verbose, so I avoided it for this simple sample. The hard coded date should be good enough to keep the query quick enough and signal to the developer that they should include a timestamp in the query

lesv (approved):

PR #335, opened 2020-11-13
chore(samples): fix version replacement tags

chingor13:

### DIFF
--- a/samples/install-without-bom/pom.xml
+++ b/samples/install-without-bom/pom.xml
@@ -23,8 +23,8 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
   </properties>
 
-  <!-- {x-version-update-start:google-cloud-logging:released} -->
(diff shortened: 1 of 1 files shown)


Daniel-Sanche:
Hmm when I cd into the samples directory and run 
```
mvn -B \
          -Penable-samples \
          -DtrimStackTrace=false \
          -Dclirr.skip=true \
          -Denforcer.skip=true \
          -fae \
          verify
```

It still fails for me. This seems to match the behaviour of the [nightly test](https://g3c.corp.google.com/results/invocations/8b34fabc-74a6-4799-a1dd-34513269c151/targets/cloud-devrel%2Fclient-libraries%2Fjava%2Fjava-logging%2Fnightly%2Fsamples/log). I'm curious why the presubmit samples test is able to pass when the nightly samples test fails

daniel-sanche (approved):

PR #380, opened 2020-12-17
fix: 2.0 has been released

elharo:
@chingor13


chingor13:
This was included in #381

PR #513, opened 2021-04-28
Make MonitoredResourceUtil#getValue not throw NPE on missing/malformed values

### LINKED ISSUE #512: MonitoredResourceUtil#getResource non-uniformly throws exceptions on missing/malformed values
I use `MonitoredResourceUtil#getResource` to load resource in our log4j2 appender implementation (basically migrated logback implementation). Container in which application is running is missing some of the expected labels for "k8s_container" resource type, namely the zone is missing which leads to `getCloudRunLocation` to throw NPE.

Shouldn't `getResource` behave lenient against this kind of problems and try to initialize and work with as much data as possible? Especially considering that missing most of the labels will indeed not cause any issues - it's only some of them that lead to exceptions.

artem-karpenko:
Fixes #512 ☕️


artem-karpenko:
Closed due to inability to amend commit

PR #514, opened 2021-04-28
Make MonitoredResourceUtil#getValue not throw NPE on missing/malformed values

### LINKED ISSUE #512: MonitoredResourceUtil#getResource non-uniformly throws exceptions on missing/malformed values

artem-karpenko:
Fixes #512 ☕️

### DIFF
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
@@ -160,7 +160,11 @@ public class MonitoredResourceUtil {
       case ContainerName:
         if (resourceType.equals("k8s_container")) {
           String hostName = System.getenv("HOSTNAME");
-          value = hostName.substring(0, hostName.indexOf("-"));
+          if (hostName != null && hostName.contains("-")) {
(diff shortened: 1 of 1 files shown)


simonz130 (approved):

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -188,7 +192,7 @@ private static String getValue(Label label, String resourceType) {
         try {
           value = new String(Files.readAllBytes(Paths.get(filePath)), StandardCharsets.UTF_8);
         } catch (IOException e) {
-          throw new LoggingException(e, true);
+          value = null;
```
simonz130, line 195: This needs to have a test

simonz130 (changes requested):

PR #533, opened 2021-05-19
feat: add cloud function autodetection

simonz130:
- Adds Cloud Function auto-detection. Currently relies on environment variables

### DIFF
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
@@ -44,6 +44,7 @@ public class MonitoredResourceUtil {
     ContainerName("container_name"),
     InstanceId("instance_id"),
     InstanceName("instance_name"),
+    Function_Name("function_name"),
     Location("location"),
     ModuleId("module_id"),
     NamespaceId("namespace_id"),
@@ -54,6 +55,7 @@ public class MonitoredResourceUtil {
     RevisionName("revision_name"),
     ServiceName("service_name"),
     VersionId("version_id"),
+    Region("region"),
     Zone("zone");
 
     private final String key;
@@ -74,6 +76,7 @@ public class MonitoredResourceUtil {
     GaeAppStandard("gae_app_standard"),
     GceInstance("gce_instance"),
     K8sContainer("k8s_container"),
+    CloudFunction("cloud_function"),
     Global("global");
 
     private final String key;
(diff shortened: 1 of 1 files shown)


google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -216,6 +227,12 @@ private static String getValue(Label label, String resourceType) {
 
   /* Detect monitored Resource type using environment variables, else return global as default. */
   private static Resource getAutoDetectedResourceType() {
+
+    // TODO detect the function from metadata server. If not available - use env var
+    if (System.getenv("FUNCTION_NAME") != null) {
```
minherz, line 232: I see no use of the environment variable's value. Does it intended to be used as a flag? In that case the name should follow a different convention.
Leaving the behavior in the hanging state isn't good solution. Let's return "global" until we determine the metadata server attribute.
Also, introducing a logic based on hard-coded environment variables is not a best strategy. A variable can be override by the function's logic or, if it is used as a fallback scenario, lead to a different behavior in real and simulated environments.

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -207,6 +214,10 @@ private static String getValue(Label label, String resourceType) {
       case Zone:
         value = MetadataConfig.getZone();
         break;
+      case Region:
+        Object region = MetadataConfig.getAttribute("instance/region");
```
minherz, line 218: Should we implement `MetadataConfig.getRegion()`?

minherz (changes requested):

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -207,6 +214,10 @@ private static String getValue(Label label, String resourceType) {
...
+        value = (region != null) ? region.toString() : "";
+
```
minherz, line 220: need `break;`

minherz:
@googlebot I consent.

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
minherz: The public documentation does not describe it but a quick test discover the following environment variables set (for golang environment): `K_REVISION`, `GCF_BLOCK_RUNTIME_go112`, `GCF_BLOCK_RUNTIME_nodejs6`, `K_SERVICE`, `FUNCTION_TARGET`, `FUNCTION_SIGNATURE_TYPE`. There is no `FUNCTION_NAME` environment variable.

PR #539, opened 2021-05-24
test: add env-tests-logging submodule

simonz130:
Adding environment tests repo as a submodule.

This PR is part of onboarding to environment tests framework for java-logging repo.


daniel-sanche (approved):
nit: I would probably name the folder something like "environment-tests" instead of "env-tests-logging". But LGTM as is

PR #602, opened 2021-07-30
feat: implement listLogs API and provide sample snippet

### LINKED ISSUE #358: Add sample to list available logs
Similar to this [Nodejs sample](https://cloud.google.com/logging/docs/samples/logging-list-logs), add a Java sample showcasing getting available logs & accompanying test(s).

Region tag: logging_list_logs

### LINKED ISSUE #593: feat: Implement user friendly interface for listLogs API
It is hard and contra-intuitive to use auto-generated implementation of the grpc or REST APIs. The Java logging client library is missing [list logs](https://cloud.google.com/logging/docs/reference/v2/rpc/google.logging.v2#google.logging.v2.LoggingServiceV2.ListLogs) implementation.

I would like to have `listLogs` interface implemented in a similar way as other interfaces (e.g. listSinks).

**Describe alternatives you've considered**
An alternative would be to use auto-generated API classes such as com.google.logging.v2.ListLogsRequest.

**Additional context**
None

minherz:
Implements listLogs interface for com.google.cloud.logging.Logging interface.
Implements unit testing for the listLogs interface.
Adds sample snippet for listLogs interface.
Refactors com.example.logging.ListLogs to incorporate listLogs snippet into the class.

Fixes #593 
Fixes #358

### DIFF
--- a/google-cloud-logging/clirr-ignored-differences.xml
+++ b/google-cloud-logging/clirr-ignored-differences.xml
@@ -6,12 +6,12 @@
     <difference>
         <differenceType>7012</differenceType>
         <className>com/google/cloud/logging/Logging</className>
-        <method>*</method>
+        <method>* listLogs*(com.google.cloud.logging.Logging$ListOption[])</method>
     </difference>
     <!-- Added methods to com.google.cloud.logging.spi.v2.LoggingRpc interface with implementation in GrpcLoggingRpc is always okay -->
     <difference>
         <differenceType>7012</differenceType>
         <className>com/google/cloud/logging/spi/v2/LoggingRpc</className>
-        <method>*</method>
+        <method>* list(com.google.logging.v2.ListLogsRequest)</method>
     </difference>
 </differences>
--- a/google-cloud-logging/clirr-ignored-differences.xml
+++ b/google-cloud-logging/clirr-ignored-differences.xml
@@ -12,6 +12,6 @@
     <difference>
         <differenceType>7012</differenceType>
         <className>com/google/cloud/logging/spi/v2/LoggingRpc</className>
-        <method>* list(com.google.logging.v2.ListLogsRequest)</method>
+        <method>* listLogs(com.google.logging.v2.ListLogsRequest)</method>
     </difference>
 </differences>
--- a/google-cloud-logging/clirr-ignored-differences.xml
+++ b/google-cloud-logging/clirr-ignored-differences.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- see http://www.mojohaus.org/clirr-maven-plugin/examples/ignored-differences.html -->
+<!-- added to resolve breaking changes until JDK will be upgraded to ≥1.8 -->
+<differences>
+    <!-- Added methods to com.google.cloud.logging.Logging interface with implementation in LoggingImpl is always okay -->
+    <difference>
+        <differenceType>7012</differenceType>
+        <className>com/google/cloud/logging/Logging</className>
+        <method>*</method>
+    </difference>
+    <!-- Added methods to com.google.cloud.logging.spi.v2.LoggingRpc interface with implementation in GrpcLoggingRpc is always okay -->
+    <difference>
+        <differenceType>7012</differenceType>
+        <className>com/google/cloud/logging/spi/v2/LoggingRpc</className>
+        <method>*</method>
+    </difference>
+</differences>
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/Logging.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/Logging.java
@@ -415,6 +415,47 @@ public interface Logging extends AutoCloseable, Service<LoggingOptions> {
   ApiFuture<Boolean> deleteSinkAsync(String sink);
 
   /**
+   * Lists the log names. This method returns a {@link Page} object that can be used to consume
+   * paginated results. Use {@link ListOption} to specify the page size or the page token from which
+   * to start listing logs.
+   *
+   * <p>Example of listing log names, specifying the page size.
+   *
+   * <pre>{@code
+   * Page<Log> logNames = logging.listLogs(ListOption.pageSize(100));
+   * Iterator<Log> logIterator = logNames.iterateAll().iterator();
+   * while (logIterator.hasNext()) {
+   *   String logName = logIterator.next();
+   *   // do something with the log name
+   * }
+   * }</pre>
+   *
+   * @throws LoggingException upon failure
+   */
+  Page<String> listLogs(ListOption... options);
+
+  /**
+   * Sends a request for listing log names. This method returns a {@code ApiFuture} object to
+   * consume the result. {@link ApiFuture#get()} returns an {@link AsyncPage} object that can be
+   * used to asynchronously handle paginated results. Use {@link ListOption} to specify the page
+   * size or the page token from which to start listing log names.
+   *
+   * <p>Example of asynchronously listing log names, specifying the page size.
+   *
+   * <pre>{@code
+   * ApiFuture<AsyncPage<Log>> future = logging.listLogsAsync(ListOption.pageSize(100));
+   * // ...
+   * AsyncPage<Sink> logNames = future.get();
+   * Iterator<Sink> logIterator = logNames.iterateAll().iterator();
+   * while (logIterator.hasNext()) {
+   *   String logName = logIterator.next();
+   *   // do something with the log name
+   * }
+   * }</pre>
+   */
+  ApiFuture<AsyncPage<String>> listLogsAsync(ListOption... options);
+
+  /**
    * Deletes a log and all its log entries. The log will reappear if new entries are written to it.
    *
    * <p>Example of deleting a log.
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/Logging.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/Logging.java
@@ -432,7 +432,10 @@ public interface Logging extends AutoCloseable, Service<LoggingOptions> {
    *
    * @throws LoggingException upon failure
    */
-  Page<String> listLogs(ListOption... options);
+  default Page<String> listLogs(ListOption... options) {
+    throw new UnsupportedOperationException(
+        "method listLogs() does not have default implementation");
+  }
 
   /**
    * Sends a request for listing log names. This method returns a {@code ApiFuture} object to
@@ -453,7 +456,10 @@ public interface Logging extends AutoCloseable, Service<LoggingOptions> {
    * }
    * }</pre>
    */
-  ApiFuture<AsyncPage<String>> listLogsAsync(ListOption... options);
+  default ApiFuture<AsyncPage<String>> listLogsAsync(ListOption... options) {
+    throw new UnsupportedOperationException(
+        "method listLogsAsync() does not have default implementation");
+  }
 
   /**
    * Deletes a log and all its log entries. The log will reappear if new entries are written to it.
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
@@ -199,8 +199,6 @@ class LoggingImpl extends BaseService<LoggingOptions> implements Logging {
 
   private static class LogNamePageFetcher extends BasePageFetcher<String> {
 
-    private static final long serialVersionUID = 7156613795615829594L;
-
     LogNamePageFetcher(
         LoggingOptions serviceOptions, String cursor, Map<Option.OptionType, ?> requestOptions) {
       super(serviceOptions, cursor, requestOptions);
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
@@ -384,9 +384,8 @@ class LoggingImpl extends BaseService<LoggingOptions> implements Logging {
   /**
    * Creates a new {@code ListLogsRequest} object.
    *
-   * Builds an instance of {@code ListLogsRequest} using page size, page token and project id
-   * from the {@code LoggingOptions}.
-   * The project id is used as the request's parent parameter.
+   * <p>Builds an instance of {@code ListLogsRequest} using page size, page token and project id
+   * from the {@code LoggingOptions}. The project id is used as the request's parent parameter.
    *
    * @see com.google.logging.v2.ListLogEntriesRequest
    * @return the created {@code ListLogsRequest} object
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
@@ -65,6 +65,8 @@ import com.google.logging.v2.ListLogEntriesRequest;
 import com.google.logging.v2.ListLogEntriesResponse;
 import com.google.logging.v2.ListLogMetricsRequest;
 import com.google.logging.v2.ListLogMetricsResponse;
+import com.google.logging.v2.ListLogsRequest;
+import com.google.logging.v2.ListLogsResponse;
 import com.google.logging.v2.ListMonitoredResourceDescriptorsRequest;
 import com.google.logging.v2.ListMonitoredResourceDescriptorsResponse;
 import com.google.logging.v2.ListSinksRequest;
@@ -192,6 +194,21 @@ class LoggingImpl extends BaseService<LoggingOptions> implements Logging {
     @Override
     public ApiFuture<AsyncPage<Sink>> getNextPage() {
       return listSinksAsync(serviceOptions(), requestOptions());
+    }
+  }
+
+  private static class LogNamePageFetcher extends BasePageFetcher<String> {
+
+    private static final long serialVersionUID = 5L;
+
+    LogNamePageFetcher(
+        LoggingOptions serviceOptions, String cursor, Map<Option.OptionType, ?> requestOptions) {
+      super(serviceOptions, cursor, requestOptions);
+    }
+
+    @Override
+    public ApiFuture<AsyncPage<String>> getNextPage() {
+      return listLogsAsync(serviceOptions(), requestOptions());
     }
   }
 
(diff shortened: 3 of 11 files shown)


google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
```diff
@@ -195,6 +197,21 @@ LoggingOptions serviceOptions() {
     }
   }
 
+  private static class LogNamePageFetcher extends BasePageFetcher<String> {
+
+    private static final long serialVersionUID = 7156613795615829594L;
```
simonz130, line 202: what is this?

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
```diff
@@ -366,6 +383,54 @@ public boolean deleteSink(String sink) {
...
+        list,
+        new Function<ListLogsResponse, AsyncPage<String>>() {
+          @Override
+          public AsyncPage<String> apply(ListLogsResponse listLogsResponse) {
+            List<String> logNames =
+                listLogsResponse.getLogNamesList() == null
+                    ? ImmutableList.<String>of()
+                    : listLogsResponse.getLogNamesList();
+            String cursor =
+                listLogsResponse.getNextPageToken().equals("")
```
simonz130, line 415: what if the token returns null?

google-cloud-logging/clirr-ignored-differences.xml
```diff
@@ -0,0 +1,17 @@
...
+        <differenceType>7012</differenceType>
+        <className>com/google/cloud/logging/Logging</className>
+        <method>*</method>
+    </difference>
+    <!-- Added methods to com.google.cloud.logging.spi.v2.LoggingRpc interface with implementation in GrpcLoggingRpc is always okay -->
+    <difference>
+        <differenceType>7012</differenceType>
+        <className>com/google/cloud/logging/spi/v2/LoggingRpc</className>
+        <method>*</method>
+    </difference>
```
chingor13, line 16: We should not just add a standing ignore for this breaking change. We cannot be making breaking changes to our clients.

The libraries will migrated to Java 8+ next week and then we can implement no-op default implementations on the interfaces so this won't be necessary (an won't be a breaking change).

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
```diff
@@ -1576,6 +1617,88 @@ public void testListResourceDescriptorAsyncWithOptions()
         Iterables.toArray(page.getValues(), MonitoredResourceDescriptor.class));
   }
 
+  @Test
+  public void testListLogs() {
```
simonz130, line 1621: The test should explain what it tests. I mean, not only the fact that you test the API but also what behavior specifically is tested in the implementation. It should be either clear from the name or documented in comments. Similar comment for other tests

chingor13 (dismissed):

google-cloud-logging/clirr-ignored-differences.xml
simonz130: why is this a breaking change?

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
```diff
@@ -173,6 +178,40 @@ public LogMetric apply(MetricInfo metricInfo) {
   private LoggingRpc loggingRpcMock;
   private Logging logging;
 
+  private void configureListLogsTests(List<String> returnedList, String cursor) {
```
simonz130, line 181: to make it easy for folks not familiar with the codebase - add a comment explaining what is expected in this configuration as the end result

google-cloud-logging/clirr-ignored-differences.xml
minherz: I'd like avoid discussing whether adding a new method to an interface is a breaking change given we aren't discussing Microsoft COM interfaces.
Since this PR does not release a new version of the library, I would like to propose the following:
Once you finish upgrading to 1.8, please open P1 issue to remove the exceptions configuration and assign the issue to me. I will handle it within a day.
It will let me to avoid putting on hold this change and need for additional merges with more recent work that is already ongoing.
WDYT?

samples/snippets/src/main/java/com/example/logging/ListLogs.java
```diff
@@ -28,35 +28,64 @@
...
+    printLogNames();
+
+    /** Expects an existing Cloud Logging log name as an argument. */
+    String logName = args[0];
+
+    printLogEntries(logName);
+  }
+
+  public static final void printLogNames() {
+    // [START logging_list_logs]
```
simonz130, line 42: If you add a new region tag, you also need to ensure that the doc page that should include this snippet references this repo and this region tag. Saying as FYI - no change is needed in this code.

minherz: Thank you. It will be the next step once PR is merged and HEAD points to the commit that has the code.

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
minherz: It is just auxiliary method. I will do more refactoring on the test code latter to generalize these methods. I will comment them after that work is done.

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
minherz: I've followed the existing pattern here: simple name like this - "test functionality of list logs" just test a default behavior, same as with other methods. Special cases are also tested.
I agree that being more verbose helps. Having said that, should not it be done globally after we agree on the template?

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
minherz: AFAIK, it cannot according to the contract. The call `listLogsResponse.getNextPageToken()` just a wrapper around auto-generated property. If malformed response is received an exception will be thrown before this method can be called.

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
minherz: What do you mean? A class for paged fetching has to be created for each type of request even if they return the same type of data.
All these classes have `serialVersionUID`, so the new class has it as well.

google-cloud-logging/clirr-ignored-differences.xml
minherz: We wait until next Thursday (August 5th) to see the upgrade to Java 1.8 is rolled in, so we can remove the exclusion configuration and to resolve the cliff plugin error using default implementation.

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
minherz: LMK if the pattern should be changed.

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
minherz: LMK if it is satisfying and I can resolve the conv

google-cloud-logging/clirr-ignored-differences.xml
minherz: @chingor13 I see similar exclusion list is already used in the library (see [proto-google-cloud-logging-v2/clirr-ignored-differences.xml](https://github.com/googleapis/java-logging/blob/minherz/create-list-logs/proto-google-cloud-logging-v2/clirr-ignored-differences.xml)). Given that the discussed method for resolving clirr errors after moving the repo to Java 8 did not work, would it be possible to get OK for this part?
I am changing the config to provide fine grained exception list on the method level.

google-cloud-logging/src/main/java/com/google/cloud/logging/spi/v2/LoggingRpc.java
```diff
@@ -93,6 +95,18 @@
...
   ApiFuture<Empty> delete(DeleteSinkRequest request);
 
+  /**
+   * Sends a request to list the log names in a project. This method returns a {@code ApiFuture}
+   * object to consume the result. {@link ApiFuture#get()} returns a response object containing the
+   * listing result.
+   *
+   * @param request the request object containing all of the parameters for the API call
+   */
+  default ApiFuture<ListLogsResponse> list(ListLogsRequest request) {
```
chingor13, line 105: This should probably be renamed to `listLogs` or `listLogNames`. There's already a `list()` method which is poorly named for listing entries. This now becomes an overload of the same method name with different functionality.

samples/snippets/src/main/java/com/example/logging/ListLogs.java
```diff
@@ -28,35 +28,64 @@
...
+
+    printLogNames();
+
+    /** Expects an existing Cloud Logging log name as an argument. */
+    String logName = args[0];
+
+    printLogEntries(logName);
+  }
+
+  public static final void printLogNames() {
```
chingor13, line 41: The [samples format guidelines](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/SAMPLE_FORMAT.md#class-structure) say one snippet per file.

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
```diff
@@ -1576,6 +1617,88 @@ public void testListResourceDescriptorAsyncWithOptions()
...
+  }
+
+  @Test
+  public void testListLogsNextPage() throws ExecutionException, InterruptedException {
+    EasyMock.replay(rpcFactoryMock);
+    logging = options.getService();
+    List<String> logNames1 = ImmutableList.of(LOG_NAME1, LOG_NAME2);
+    List<String> logNames2 = ImmutableList.of(LOG_NAME1);
+    String nextPageCursor = "nextCursor";
+    configureListLogsTests(logNames1, logNames2, LOG_NAMES_CURSOR, nextPageCursor);
```
chingor13, line 1651: Readability-wise, it's hard to tell what these setup methods are doing (especially since it's an overloaded method).

chingor13 (changes requested):
Feature looks fine to me (aside from the method name on the Rpc interface/implementation).

New samples need to follow the rubric/format set by the Java samples working group: https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/SAMPLE_FORMAT.md. If that is significant work, consider breaking apart the feature PR and the sample PR

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
```diff
@@ -366,6 +381,54 @@ public boolean deleteSink(String sink) {
     return transform(rpc.delete(request), EMPTY_TO_BOOLEAN_FUNCTION);
   }
 
+  private static ListLogsRequest listLogsRequest(
```
simonz130, line 384: please document the purpose of this

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
simonz130: You can add a comment about this for future code readers.

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
```diff
@@ -366,6 +381,54 @@ public boolean deleteSink(String sink) {
...
+    if (pageSize != null) {
+      builder.setPageSize(pageSize);
+    }
+    if (pageToken != null) {
+      builder.setPageToken(pageToken);
+    }
+    return builder.build();
+  }
+
+  private static ApiFuture<AsyncPage<String>> listLogsAsync(
```
simonz130, line 399: even though it's not public - please document it for the sake of future code maintainers

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
simonz130, line 181: please document both overloads of configureListLogsTests so the reader doesn't need to scroll into the implementation and figure out the differences by reading code in both methods.

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
simonz130: A lot of this codebase (tests and library) was inherited. Being more verbose helps in maintenance. We should prioritize it when possible. Changing it across the entire codebase is desirable, but I don't want to hold your PR for something that's not related. If anything, we can open a low prio technical debt bug to improve it.

simonz130 (approved):
Great work!

google-cloud-logging/clirr-ignored-differences.xml
```diff
@@ -0,0 +1,17 @@
...
+        <differenceType>7012</differenceType>
+        <className>com/google/cloud/logging/Logging</className>
+        <method>* listLogs*(com.google.cloud.logging.Logging$ListOption[])</method>
+    </difference>
+    <!-- Added methods to com.google.cloud.logging.spi.v2.LoggingRpc interface with implementation in GrpcLoggingRpc is always okay -->
+    <difference>
+        <differenceType>7012</differenceType>
+        <className>com/google/cloud/logging/spi/v2/LoggingRpc</className>
+        <method>* list(com.google.logging.v2.ListLogsRequest)</method>
+    </difference>
```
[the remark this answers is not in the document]
minherz: Unfortunately the `clirr` plugin yields error on default implementations. So, the opt out config is necessary. I've tried to narrow the scope of the opt out to the new methods only.
There is alternative plugin: `japicmp` that is expected to do the same task.

google-cloud-logging/src/main/java/com/google/cloud/logging/spi/v2/LoggingRpc.java
minherz: I've tried to followed the existing pattern. However, I will rename the method to be `listLogs`.

google-cloud-logging/clirr-ignored-differences.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- see http://www.mojohaus.org/clirr-maven-plugin/examples/ignored-differences.html -->
+<!-- added to resolve breaking changes until JDK will be upgraded to ≥1.8 -->
```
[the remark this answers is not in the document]
minherz: Yes. Without it the verification fail on the `clirr` plugin with error 7012.

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
minherz: I will document the method and will try to improve the method name too :+1:

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
minherz: Will be done

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
minherz: I cannot do this. I am not familiar with the pattern that is used to implement paging. I will look for more information about it.

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
minherz: done

samples/snippets/src/main/java/com/example/logging/ListLogs.java
minherz: The samples are broken into ListLogEntries.java for listing log entries and ListLogs.java for listing log names.

minherz:
@chingor13 i did not find the reason why the snippet-bot fails. the region tags in ListLogs.java and ListLogEntries.java match and I am unaware of other tests that the bot may perform. maybe it is due to the split of the snippets.
i tried to restart the check from the Checks console, but it either does not work or has no effect.

chingor13:
> @chingor13 i did not find the reason why the snippet-bot fails. the region tags in ListLogs.java and ListLogEntries.java match and I am unaware of other tests that the bot may perform. maybe it is due to the split of the snippets.
> i tried to restart the check from the Checks console, but it either does not work or has no effect.

The way snippets work on Cloudsite is the reference the snippet by GitHub repo, file path, and snippet name. By moving the snippet to a different file, it will break the sample usage. Moving a snippet to a new file or repo involves 3 steps:

1. Duplicate the snippet to the new file
2. Fix references to that snippet to the new location
3. Remove the original snippet.

chingor13 (changes requested):
nit on license header in new file

Otherwise make sure we're not breaking the snippets

samples/snippets/src/main/java/com/example/logging/ListLogEntries.java
```diff
@@ -0,0 +1,48 @@
+/*
+ * Copyright 2017 Google Inc.
```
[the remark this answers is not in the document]
minherz: Done. I've updated the Copyright year in all files that i've touched.

lesv (changes requested):
We do not bump copyrights.
Please ping me if you want details.

samples/snippets/src/main/java/com/example/logging/ListLogs.java
```diff
@@ -16,47 +16,34 @@
...
 }
+// [END logging_list_logs]
+
+// the following is a temporary location due to exclusive requirement
+// from snippet layout vs. snippet-bot
+
+// [START logging_list_log_entries]
+
+
+// [END logging_list_log_entries]
```
chingor13, line 49: If Cloud site is redeployed, the snippet will empty on cloud.google.com which is probably worse than the page failing to deploy.

chingor13 (dismissed):

lesv (approved):

samples/snippets/src/main/java/com/example/logging/ListLogs.java
minherz: Jeff, it is a problem of the chicken and egg. Your review request was to have one snippet per file (by guidelines). Snippet management team requires the migration to be done by first introducing the code and then moving the region tag (spoke with Takashi). It is impossible to do both at once. This PR is stuck for a week just because we discuss this subject. Please, release it. I will issue the fixing PR within an hour that will clean up the duplicated region tag.
The guidelines allow a short period of time when the particular sample will be empty. I just need the help to take the work forward.

lesv:
```
+// [START logging_list_log_entries]
+
+
+// [END logging_list_log_entries]
```
>Jeff, it is a problem of the chicken and egg. Your review request was to have one snippet per file (by guidelines). Snippet management team requires the migration to be done by first introducing the code and then moving the region tag (spoke with Takashi). It is impossible to do both at once. This PR is stuck for a week just because we discuss this subject. Please, release it. I will issue the fixing PR within an hour that will clean up the duplicated region tag.
>The guidelines allow a short period of time when the particular sample will be empty. I just need the help to take the work forward.

Woa - this seems wrong.  Where are you hearing about this process -- I'd like to fix it.

lesv:
Note - my LGTM was only for the copyrights.  Let me take another look.

lesv (approved):

google-cloud-logging/src/main/java/com/google/cloud/logging/Logging.java
```diff
@@ -414,6 +414,53 @@ public static EntryListOption folder(String folder) {
...
+   *   // do something with the log name
+   * }
+   * }</pre>
+   *
+   * @throws LoggingException upon failure
+   */
+  default Page<String> listLogs(ListOption... options) {
+    throw new UnsupportedOperationException(
+        "method listLogs() does not have default implementation");
+  }
```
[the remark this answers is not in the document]
minherz: @lesv can you ref me to guideline about providing (or not) examples in javadoc?

lesv:
In thinking about it, they are trying to create an interface with default
methods here.  (a Java 8 feature).  Since *logging* will never be created
directly, you'll never execute this code.  I believe I closed the comment
before I LGTM'd the review.  Sorry it was sent.

PR #615, opened 2021-08-11
fix: add space to fix embedding exclusion in docs

simonz130:
When the bom instructions feature at https://cloud.google.com/logging/docs/reference/libraries?hl=en
there is a weird part there:
```xml
<!-- ...
</dependencies>
```

Comparing to other libraries in java, it looks like we need to add a line break.

### DIFF
--- a/samples/snippets/pom.xml
+++ b/samples/snippets/pom.xml
@@ -41,6 +41,7 @@
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-logging</artifactId>
     </dependency>
+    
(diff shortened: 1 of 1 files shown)


minherz (commented):
Should we add an empty line between lines 57 and 58 too:
```xml
</dependency>

<!-- [END_EXCLUDE] -->
</dependencies>
```
or between 58 and 58:
```xml
</dependency>
<!-- [END_EXCLUDE] -->

</dependencies>
```

?

daniel-sanche (approved):

PR #621, opened 2021-08-12
chore(doc): remove duplicated region tag for logging_list_log_entries

minherz:
Clean up the duplicated region tag logging_list_log_entries that was introduced by #602 in order to migrate the region tag into the new file


minherz:
The branch was not created from master and contains incorrect commits.

PR #627, opened 2021-08-17
feat: add support for tailLogEntry API

minherz:
Implements tailLogEntry API that exposes entries.tail API for end users. The API is in Beta yet and does not support resuming and retrying mechanisms.

Fixes #592

PR #628, opened 2021-08-17
feat: Add a sample snippet for use of the tailLogEntries API

minherz:
Implements samples for tailLogEntries API implemented in PR #627

### DIFF
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/JavaTimeConversions.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/JavaTimeConversions.java
@@ -16,6 +16,7 @@
 
 package com.google.cloud.logging;
 
+import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.math.LongMath.checkedAdd;
 import static com.google.common.math.LongMath.checkedSubtract;
 
@@ -33,8 +34,8 @@ import java.time.Instant;
  * The class complements convertion methods that are currently not supported in the published
  * protobuf-java-util. After migrating protobuf-java-util to Java 8 this class can be removed.
  *
- * @see <a
- *     href="https://mvnrepository.com/artifact/com.google.protobuf/protobuf-java-util">protobuf-java-util</a>
+ * @see <a href=
+ *     "https://mvnrepository.com/artifact/com.google.protobuf/protobuf-java-util">protobuf-java-util</a>
  */
 class JavaTimeConversions {
   static final long NANOS_PER_SECOND = 1000000000;
@@ -48,6 +49,7 @@ class JavaTimeConversions {
    *     Timestamps#isValid}.
    */
   public static Instant toJavaInstant(Timestamp timestamp) {
+    checkNotNull(timestamp);
     timestamp = normalizedTimestamp(timestamp.getSeconds(), timestamp.getNanos());
     return Instant.ofEpochSecond(timestamp.getSeconds(), timestamp.getNanos());
   }
@@ -59,6 +61,7 @@ class JavaTimeConversions {
    *     a {@link Timestamp}. See {@link Timestamps#isValid}.
    */
   public static Timestamp toProtoTimestamp(Instant instant) {
+    checkNotNull(instant);
     return normalizedTimestamp(instant.getEpochSecond(), instant.getNano());
   }
 
--- a/google-cloud-logging/src/test/java/com/google/cloud/logging/it/ITTailLogsTest.java
+++ b/google-cloud-logging/src/test/java/com/google/cloud/logging/it/ITTailLogsTest.java
@@ -34,10 +34,6 @@ public class ITTailLogsTest extends BaseSystemTest {
   private static final MonitoredResource GLOBAL_RESOURCE =
       MonitoredResource.newBuilder("global").build();
 
-  // @BeforeClass
-  // public static void configureWriteLogs() {
-  // }
-
   @AfterClass
   public static void cleanUpLogs() throws InterruptedException {
     assertTrue(cleanupLog(LOG_ID));
@@ -83,7 +79,9 @@ public class ITTailLogsTest extends BaseSystemTest {
     assertEquals(testLogEntry.getLogName(), resultEntry.getLogName());
     assertEquals(
         testLogEntry.getHttpRequest().getStatus(), resultEntry.getHttpRequest().getStatus());
+    // assert equals for Payload objects
     assertTrue(testLogEntry.getPayload().equals(resultEntry.getPayload()));
+    // assert equals for Map<String,String> objects
     assertTrue(testLogEntry.getLabels().equals(resultEntry.getLabels()));
   }
 }
--- a/samples/snippets/src/main/java/com/example/logging/TailLogEntries.java
+++ b/samples/snippets/src/main/java/com/example/logging/TailLogEntries.java
@@ -1,0 +1,21 @@
+/*
+ * Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+public class TailLogEntries {
+
+    public static void main(String[] args) throws Exception {
+    }
+}
--- a/samples/snippets/src/main/java/com/example/logging/TailLogEntries.java
+++ b/samples/snippets/src/main/java/com/example/logging/TailLogEntries.java
@@ -14,8 +14,41 @@
  * limitations under the License.
  */
 
+package com.example.logging;
+
+// [START logging_write_log_entry]
+import com.google.cloud.logging.LogEntry;
+import com.google.cloud.logging.LogEntryServerStream;
+import com.google.cloud.logging.Logging;
+import com.google.cloud.logging.Logging.TailOption;
+import com.google.cloud.logging.LoggingOptions;
+
 public class TailLogEntries {
 
-    public static void main(String[] args) throws Exception {
+  public static void main(String[] args) throws Exception {
+    // TODO(developer): Optionally provide the logname as an argument.
+    String logName = args.length > 0 ? args[0] : "";
+
+    LoggingOptions options = LoggingOptions.getDefaultInstance();
+    try (Logging logging = options.getService()) {
+
+      // Optionally compose a filter to tail log entries only from specific log
+      LogEntryServerStream stream;
+
+      if (logName != "") {
+        stream =
+            logging.tailLogEntries(
+                TailOption.filter(
+                    "logName=projects/" + options.getProjectId() + "/logs/" + logName));
+      } else {
+        stream = logging.tailLogEntries();
+      }
+      System.out.println("start streaming..");
+      for (LogEntry log : stream) {
+        System.out.println(logEntry);
+      }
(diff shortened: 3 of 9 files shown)


google-cloud-logging/src/main/java/com/google/cloud/logging/JavaTimeConversions.java
```diff
@@ -0,0 +1,80 @@
...
+
+  /**
+   * Converts a protobuf {@link Timestamp} to a {@link java.time.Instant}.
+   *
+   * @throws IllegalArgumentException if the given {@link Timestamp} is not valid. See {@link
+   *     Timestamps#isValid}.
+   */
+  public static Instant toJavaInstant(Timestamp timestamp) {
+    timestamp = normalizedTimestamp(timestamp.getSeconds(), timestamp.getNanos());
+    return Instant.ofEpochSecond(timestamp.getSeconds(), timestamp.getNanos());
```
simonz130, line 52: timestamp is class so can be null. Add a check for null

google-cloud-logging/src/main/java/com/google/cloud/logging/JavaTimeConversions.java
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  /**
+   * Converts a {@link java.time.Instant} to a protobuf {@link Timestamp}.
+   *
+   * @throws IllegalArgumentException if the given {@link java.time.Instant} cannot legally fit into
+   *     a {@link Timestamp}. See {@link Timestamps#isValid}.
+   */
+  public static Timestamp toProtoTimestamp(Instant instant) {
+    return normalizedTimestamp(instant.getEpochSecond(), instant.getNano());
```
simonz130, line 62: ditto here

google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
```diff
@@ -133,14 +131,44 @@ public Builder setResource(MonitoredResource resource) {
     /**
      * Sets the time at which the event described by the log entry occurred, in milliseconds. If
      * omitted, the Logging service will use the time at which the log entry is received.
+     *
+     * @deprecated This method is no longer recommended to setup the entry timestamp.
+     *     <p>Use {@link setTimeStamp(Instant)} instead.
```
simonz130, line 136: nit: Timestamp (small s)

google-cloud-logging/src/test/java/com/google/cloud/logging/BaseSystemTest.java
```diff
@@ -56,17 +56,29 @@ public static void afterClass() throws Exception {
...
   protected static <V> String createEqualityFilter(String name, V value) {
     return name + "=" + "\"" + value.toString() + "\"";
   }
 
+  protected static boolean cleanupLog(String logName) throws InterruptedException {
+    int deleteAttempts = 0;
+    int allowedDeleteAttempts = 5;
+    boolean deleted = false;
+    while (!deleted && deleteAttempts < allowedDeleteAttempts) {
+      Thread.sleep(5000);
```
simonz130, line 71: add logs so when looking at the test run logs we can see how many attempts it took to clean and whether cleanup succeeded or not.

google-cloud-logging/src/test/java/com/google/cloud/logging/it/ITTailLogsTest.java
```diff
@@ -0,0 +1,89 @@
...
+            .setHttpRequest(HttpRequest.newBuilder().setStatus(200).build())
+            .setResource(GLOBAL_RESOURCE)
+            .build();
+
+    String filter = "logName=projects/" + logging.getOptions().getProjectId() + "/logs/" + LOG_ID;
+    LogEntryServerStream stream = logging.tailLogEntries(TailOption.filter(filter));
+
+    Runnable task =
+        () -> {
+          // it may take awhile for tailing session to startup on the backend
```
simonz130, line 61: is it documented somewhere on cloud.google.com (how long it might take)? If not, worth opening a bug on docs team to add this - I find it non-trivial assumption that customers of this API might get confused with

google-cloud-logging/src/test/java/com/google/cloud/logging/it/ITTailLogsTest.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+    stream.cancel();
+
+    LogEntry resultEntry = receivedEntries.get(0);
+    assertEquals(testLogEntry.getResource().getType(), resultEntry.getResource().getType());
+    assertEquals(testLogEntry.getLogName(), resultEntry.getLogName());
+    assertEquals(
+        testLogEntry.getHttpRequest().getStatus(), resultEntry.getHttpRequest().getStatus());
+    assertTrue(testLogEntry.getPayload().equals(resultEntry.getPayload()));
+    assertTrue(testLogEntry.getLabels().equals(resultEntry.getLabels()));
```
simonz130, line 87: For the sake of folks who might touch this code very occasionally, it's worth adding a brief comment what you are testing here.

google-cloud-logging/src/test/java/com/google/cloud/logging/it/ITTailLogsTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+public class ITTailLogsTest extends BaseSystemTest {
+
+  private static final String LOG_ID = formatForTest("test-tail-log-entries-log");
+  private static final MonitoredResource GLOBAL_RESOURCE =
+      MonitoredResource.newBuilder("global").build();
+
+  // @BeforeClass
+  // public static void configureWriteLogs() {
+  // }
```
simonz130, line 39: remove this

google-cloud-logging/src/test/java/com/google/cloud/logging/it/ITTailLogsTest.java
```diff
@@ -0,0 +1,89 @@
...
+          // wait 10 sec before sending log entries
+          try {
+            Thread.sleep(10000);
+          } catch (InterruptedException t) {
+          }
+          logging.write(ImmutableList.of(testLogEntry));
+        };
+    Thread thread = new Thread(task);
+    thread.start();
+
```
simonz130, line 71: don't you need to do thread.join here to wait in the main thread until the write part will complete or will the loop in line 73 handle that implicitly (stream won't return anything so iterator will block until there is anything to return?)

google-cloud-logging/src/test/java/com/google/cloud/logging/it/ITTailLogsTest.java
```diff
@@ -0,0 +1,89 @@
...
+    final ArrayList<LogEntry> receivedEntries = new ArrayList<>();
+    for (LogEntry log : stream) {
+      receivedEntries.add(log);
+      if (receivedEntries.size() > 0) {
+        break;
+      }
+    }
+    stream.cancel();
```
simonz130, line 79: cancel or close?

simonz130 (changes requested):
Great job. Please rebase it and look into my question on thread.join

google-cloud-logging/src/main/java/com/google/cloud/logging/JavaTimeConversions.java
minherz: :+1:

google-cloud-logging/src/main/java/com/google/cloud/logging/JavaTimeConversions.java
minherz: :+1:

google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
minherz: Yes, because the "timestamp" here is not a reference to the type.

google-cloud-logging/src/test/java/com/google/cloud/logging/it/ITTailLogsTest.java
minherz: Not exactly. The public doc states that it starts streaming new log entries but it does not define how quick it starts doing it. The delay might reach few seconds. The 10 second is placed to avoid flakiness in the test.
Mind that this API is still in Beta. They are working to improve the stream establishing time.

google-cloud-logging/src/test/java/com/google/cloud/logging/it/ITTailLogsTest.java
minherz: :+:

google-cloud-logging/src/test/java/com/google/cloud/logging/it/ITTailLogsTest.java
minherz: No, there is no need to wait, quite an opposite. The response stream is being established when the code starts waiting for next response to come. So, it will be waste of time to wait on join. The main thread of the test is guaranteed to finish only after the first item is received.

google-cloud-logging/src/test/java/com/google/cloud/logging/it/ITTailLogsTest.java
minherz: The convention is to call `cancel()`. I am not sure the class exposes `close()`.

google-cloud-logging/src/test/java/com/google/cloud/logging/it/ITTailLogsTest.java
minherz: I would say that the code is self explanatory. The names of variables as well as names of methods.
I would refactor the two last assert validations to be `assertEquals` as well since this is what essentially is done here.

google-cloud-logging/src/test/java/com/google/cloud/logging/BaseSystemTest.java
minherz: I will do it later. Need review what and how it is better done, unless you are sure that a plain `System.out.println()` call will be good enough. I just moved the existing code into the base class. Adding this info is useful but might have few gotchas.

samples/snippets/src/test/java/com/example/logging/LoggingIT.java
```diff
@@ -135,4 +134,36 @@ public void testListLogNames_shouldPass() throws Exception {
...
+              LogEntry logEntry =
+                  LogEntry.newBuilder(StringPayload.of(STRING_PAYLOAD))
+                      .setLogName(TEST_LOG)
+                      .setResource(MonitoredResource.newBuilder("global").build())
+                      .build();
+              // Writes the log entry asynchronously
+              logging.write(Collections.singleton(logEntry));
+            }
+          } catch (Exception t) {
+          }
```
simonz130: nit: something weird here with indentation - worth re-aligning so it's clearer which '{' belongs to which '}'

samples/snippets/src/main/java/com/example/logging/TailLogEntries.java
```diff
@@ -0,0 +1,55 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example.logging;
+
+// [START logging_write_log_entry]
```
simonz130, line 19: This should be **logging_tail_log_entries**

samples/snippets/src/main/java/com/example/logging/TailLogEntries.java
```diff
@@ -0,0 +1,55 @@
...
+      System.out.println("start streaming..");
+      for (LogEntry log : stream) {
+        System.out.println(log);
+        // cancel infinite streaming after receiving first entry
+        stream.cancel();
+      }
+    }
+  }
+}
+// [END logging_write_log_entry]
```
simonz130, line 55: ditto logging_tail_log_entries

simonz130 (approved):

PR #642, opened 2021-08-25
fix: Temporary "fix How containername is determined from hostname

### LINKED ISSUE #494: Issue with container name as reported by k8s_container resource type
When using java-logging via [java-logging-logback](https://github.com/googleapis/java-logging-logback) on GKE, logs emitted from applications do not contain the correct container name. This was reported in https://github.com/googleapis/java-logging/issues/206#issuecomment-774383829 after support for k8s_container was first merged.

The underlying issue is here: https://github.com/googleapis/java-logging/pull/207/files#diff-aeaea0f21b13f50ea525086b8cb13b790db972e6f610620c5b9d2a17768821beR162-R163, where the `HOSTNAME` environment variable is used to determine the container name. This is incorrect for a number of reasons. First, the hostname may have no relation to the container name; by default it is the pod name, but can even be overridden to be something else. Second, the code splits the hostname on dashes assuming that there are no dashes in the pod name, so if you have dashes in your pod name it just gives you the first piece.

I'm not sure what a better solution to this would be. I couldn't find a way to expose the container name in a reliable name like I was able to do for the namespace name by using `downwardAPI` and setting `KUBERNETES_NAMESPACE_FILE`. I'm willing to try and tackle fixing this but I'd like to have a discussion around the best way to get the container name.

Relatedly, the reported pod name will be the `HOSTNAME` as well. This may work so long as the hostname is not overridden in the pod spec, but if it is overridden then this value will be incorrect. The `downwardAPI` approach I used to get the namespace (instead of requiring that `automountServiceAccountToken` be set to `true`) would work, but the pod spec gets quite gross.

wildhemp:
The container name can contain '-' character, so String.indexOf is not a good way to determine it from hostname. Instead it should be substring(0, <second to last '-'>), which is what this fix is doing.

Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/java-logging/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [ ] Ensure the tests and linter pass
- [ ] Code coverage does not decrease (if any source code was changed)
- [ ] Appropriate docs were updated (if necessary)

Fixes #494


simonz130 (approved):

simonz130:
@wildhemp - there were github changes to ubuntu image that affected our CI. As a result, your PR failed on CI. I mitigated it by updating the CI but now you need to sync your form with latest changes to our fork master and then your PR will pass the CI and I'll merge it.

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -160,7 +160,7 @@ private static String getValue(Label label, String resourceType) {
       case ContainerName:
         if (resourceType.equals("k8s_container")) {
           String hostName = System.getenv("HOSTNAME");
-          value = hostName.substring(0, hostName.indexOf("-"));
+          value = hostName.substring(0, hostName.lastIndexOf("-", hostName.lastIndexOf("-") - 1));
```
minherz, line 163: Need unit testing for corner cases.

minherz (changes requested):

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
wildhemp: This is not too easy to unit test, because everything is done via static methods and the hostname is read using System.getenv, which is brittle to fiddle with (probably that's the reason there are no really unit tests even for making sure the current code works.)

One thing I can do to test for edge cases would be to pull out the 

value = hostName.substring(0, hostName.lastIndexOf("-", hostName.lastIndexOf("-") - 1));

to a utility class (K8sContainerUtil?) and test only this one call via a unit test. Would that be acceptable? Or you have any better idea on how to go about testing this.

simonz130: Yes, I think it works. Thanks for your contribution, @wildhemp

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -160,7 +164,11 @@ private static String getValue(Label label, String resourceType) {
       case ContainerName:
         if (resourceType.equals("k8s_container")) {
           String hostName = System.getenv("HOSTNAME");
-          value = hostName.substring(0, hostName.indexOf("-"));
+          if (hostName == null) {
+            value = "unknown";
```
simonz130, line 166: This logic should be reverted on its head: first asking metadata endpoint, then if exception, null or empty => ask env variable (HOSTNAME), then trying to figure out the container name from host name. Also probably for better readability would be better to just pass the value of hostName and make the determination of the value inside of getContainerNameFromHostName. This way there are no two places for the logic.

google-cloud-logging/src/test/java/com/google/cloud/logging/K8sContainerUtilTest.java
```diff
@@ -0,0 +1,33 @@
...
+import static com.google.common.truth.Truth.assertThat;
+
+public class K8sContainerUtilTest {
+
+  @Test
+  public void testGetContainerNameProperlyFormatted() {
+    assertThat(getContainerNameFromHostName("test-one-68449bf55d-44pdr"))
+        .isEqualTo("test-one");
+    assertThat(getContainerNameFromHostName("test-again-d667d8f5-rlxz5"))
+        .isEqualTo("test-again");
```
simonz130: is there a difference between these 2 usecases?

google-cloud-logging/src/test/java/com/google/cloud/logging/K8sContainerUtilTest.java
```diff
@@ -0,0 +1,33 @@
...
+    assertThat(getContainerNameFromHostName("test-again-again-again-d667d8f5-rlxz5"))
+        .isEqualTo("test-again-again-again");
+    assertThat(getContainerNameFromHostName("test-test--44pdr"))
+        .isEqualTo("test-test");
+  }
+
+  @Test
+  public void testGetContainerNameWrongFormat() {
+    assertThat(getContainerNameFromHostName("test-one-68449bf55d-68449bf55d"))
+        .isEqualTo("test-one-68449bf55d-68449bf55d");
```
simonz130, line 38: what about values like x--x, x?

google-cloud-logging/src/test/java/com/google/cloud/logging/K8sContainerUtilTest.java
wildhemp: Just different length in the deployment id.

google-cloud-logging/src/test/java/com/google/cloud/logging/K8sContainerUtilTest.java
wildhemp: that would be x--x. Added a test.

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
wildhemp: I'm asking metadata but afaics it just returns null if there's any exception, so I'm checking that instead of catching exception. Also, I'm not just blanket getting the hostname, but only when it's a k8s_container.
Alternatively I could just return the hostname if not k8s_container and do the logic in K8sContainerUtil if it is a k8s. wdyt?

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -158,11 +160,9 @@ private static String getValue(Label label, String resourceType) {
...
         break;
       case ContainerName:
-        if (resourceType.equals("k8s_container")) {
-          String hostName = System.getenv("HOSTNAME");
-          value = hostName.substring(0, hostName.indexOf("-"));
-        } else {
-          value = MetadataConfig.getContainerName();
+        value = MetadataConfig.getContainerName();
+        if (value == null && resourceType.equals("k8s_container")) {
+          value = getContainerNameFromHostName(System.getenv("HOSTNAME"));
```
minherz, line 165: This code retrieves **pod** name and not container name. By default containers receive the same hostname which is a [pod name](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-hostname-and-subdomain-fields) if the pod is launched as standalone or the name is generated following the template `REPLICASET-?????` where `REPLICASET` is a name of the replicaset build as `PODNAME-HASH` and `?????` are 5 alphanumeric identifier.
So, when you extract the first part of the hostname you get the pod name and not container name.

google-cloud-logging/src/main/java/com/google/cloud/logging/K8sContainerUtil.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+    int lastDashIndex = hostName.length() - 6;
+    if (lastDashIndex > 0 && hostName.charAt(lastDashIndex) == '-') {
+      int secondToLastIndex = hostName.lastIndexOf("-", lastDashIndex - 1);
+      if (secondToLastIndex > 0) {
+        return hostName.substring(0, secondToLastIndex);
+      }
+    }
+    return hostName;
+  }
```
minherz, line 47: It is not easy to understand the implementation logic. Would it be simpler to use a regexp? For example something like:
```java
const String pattern = "([a-z0-9]*)-[a-z0-9]{9}-[a-z0-9]{5}";
Pattern r = Pattern.compile(pattern);

Matcher m = r.matcher("myservice-bc8d57467-tgcbj");
if (m.find( )) {
  System.out.println("Pod name: " + m.group(0));
} else {
  System.out.println("unknown");
}
```

minherz (changes requested):

minherz:
Closing due to lack of feedback and the fact that the same topic is handled in #708

PR #666, opened 2021-09-12
doc: add information about writing log entry modes

### LINKED ISSUE #665: doc: documentation does not describe different modes for writing log entries
Existing documentation of the Java logging library does not describe synchronous vs. asynchronous logging configuration and what effect it has on writing log entries.
Also the documentation does not include batching settings configuration guidelines in a case the asynchronous logging is used.

See also issue #432 as an example when such information would be useful.

minherz:
provide an explanation about synchronous and asynchronous modes for writing log entries and how the mode can be configured.

Fixes #665

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,6 +166,27 @@ LogEntry firstEntry = LogEntry.newBuilder(StringPayload.of("message"))
 logging.write(Collections.singleton(firstEntry));
 ```
 
+The library supports writing log entries synchronously and asynchronously.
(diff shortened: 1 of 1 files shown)


minherz:
Unfortunately, the library does not provide a way to configure batching settings. It does not make sense to change the code in order to expose batching settings now. This change will be incorporated into the planed refactoring of the existing batching implementation.

daniel-sanche (approved):
LGTM, but Simon should also probably take a look since he's more familiar with the java library than I am

simonz130 (approved):

PR #708, opened 2021-10-12
feat: implement revised version of the monitored resource type discovery and metadata population

### LINKED ISSUE #685: structured logging: revise resource type and info retrieval
If not defined, the resource field in LogEntry should be auto-populated when writing log entries to backend

Resource type and additional resource metadata should be populated similarly when logged using Logging agent.
The populated metadata labels should be per resource type and adhere to [official documentation](https://cloud.google.com/monitoring/api/resources)

minherz:
Implement more robust resource type discovery including support for Cloud Function resource. Cache results per pair of the project id and resource type.
Refactor [MonitoredResourceUtil.java ](https://github.com/googleapis/java-logging/blob/main/google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java) to allow mocking static methods that read environment variables and metadata server attributes by using `ResourceTypeEnvironmentGetter` class.
Implement unit testing of the `MonitoredResourceUtil.getResource()` for each detected environment.

Fixes #685

### DIFF
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
@@ -277,7 +277,7 @@ public class LogEntry implements Serializable {
      *
      * @see <a href="https://cloud.google.com/logging/docs/view/logs_index">Log Entries and Logs</a>
      */
-    public Builder setPayload(Payload payload) {
+    public Builder setPayload(Payload<?> payload) {
       this.payload = payload;
       return this;
     }
@@ -434,7 +434,7 @@ public class LogEntry implements Serializable {
    * @see <a href="https://cloud.google.com/logging/docs/view/logs_index">Log Entries and Logs</a>
    */
   @SuppressWarnings("unchecked")
-  public <T extends Payload> T getPayload() {
+  public <T extends Payload<?>> T getPayload() {
     return (T) payload;
   }
 
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
@@ -37,6 +37,7 @@ import java.util.Map;
 public class MonitoredResourceUtil {
 
   private static final String APPENGINE_LABEL_PREFIX = "appengine.googleapis.com/";
+  private static final String CLUSTER_NAME_ATTRIBUTE = "instance/attributes/cluster-name";
   private static final String K8S_POD_NAMESPACE_PATH =
       "/var/run/secrets/kubernetes.io/serviceaccount/namespace";
   private static final String FLEX_ENV = "flex";
@@ -188,7 +189,7 @@ public class MonitoredResourceUtil {
         && getter.getEnv("FUNCTION_TARGET") != null) {
       return Resource.CloudFunction;
     }
-    if (getter.getAttribute("instance/attributes/cluster-name") != null) {
+    if (getter.getAttribute(CLUSTER_NAME_ATTRIBUTE) != null) {
       return Resource.K8sContainer;
     }
     if (getter.getAttribute("instance/preempted") != null
@@ -216,7 +217,7 @@ public class MonitoredResourceUtil {
 
     switch (label) {
       case ClusterName:
-        value = getter.getAttribute("instance/attributes/cluster-name");
+        value = getter.getAttribute(CLUSTER_NAME_ATTRIBUTE);
         break;
       case ConfigurationName:
         value = getter.getEnv("K_CONFIGURATION");
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
@@ -18,7 +18,6 @@ package com.google.cloud.logging;
 
 import com.google.cloud.MetadataConfig;
 import com.google.cloud.MonitoredResource;
-import com.google.cloud.ServiceOptions;
 import com.google.cloud.logging.LogEntry.Builder;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableMultimap;
@@ -38,19 +37,24 @@ import java.util.Map;
  */
 public class MonitoredResourceUtil {
 
+  static String K8S_POD_NAMESPACE_PATH = "/var/run/secrets/kubernetes.io/serviceaccount/namespace";
+  static String FLEX_ENV = "flex";
+  static String STD_ENV = "standard";
+
   private enum Label {
-    AppId("app_id"),
     ClusterName("cluster_name"),
+    ConfigurationName("configuration_name"),
     ContainerName("container_name"),
+    Env("env"),
+    FunctionName("function_name"),
     InstanceId("instance_id"),
     InstanceName("instance_name"),
     Location("location"),
     ModuleId("module_id"),
-    NamespaceId("namespace_id"),
     NamespaceName("namespace_name"),
-    PodId("pod_id"),
     PodName("pod_name"),
     ProjectId("project_id"),
+    Region("region"),
     RevisionName("revision_name"),
     ServiceName("service_name"),
     VersionId("version_id"),
@@ -69,9 +73,8 @@ public class MonitoredResourceUtil {
 
   private enum Resource {
     CloudRun("cloud_run_revision"),
-    Container("container"),
-    GaeAppFlex("gae_app_flex"),
-    GaeAppStandard("gae_app_standard"),
+    CloudFunction("cloud_function"),
+    AppEngine("gae_app"),
     GceInstance("gce_instance"),
     K8sContainer("k8s_container"),
     Global("global");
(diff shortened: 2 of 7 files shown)


google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
```diff
@@ -434,7 +434,7 @@ public SourceLocation getSourceLocation() {
    * @see <a href="https://cloud.google.com/logging/docs/view/logs_index">Log Entries and Logs</a>
    */
   @SuppressWarnings("unchecked")
-  public <T extends Payload> T getPayload() {
+  public <T extends Payload<?>> T getPayload() {
```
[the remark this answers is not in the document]
minherz: No, it is a correct syntax that also eliminates warnings.

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -144,135 +206,152 @@ public static MonitoredResource getResource(String projectId, String resourceTyp
...
-    Resource resourceType = getAutoDetectedResourceType();
+    Resource resourceType = detectResourceType();
     return createEnhancers(resourceType);
   }
 
+  @SuppressWarnings("incomplete-switch")
   private static String getValue(Label label, String resourceType) {
-    String value;
+    String value = "";
+
```
simonz130, line 216: This whole method will need to be refactored (doesn't need to be in this PR). The switch...case will become really long really soon (more detected environments from 220+ list). I'll review it assuming it stays this way for now but please open a task on this repo to refactor it to be more sane.

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -144,135 +206,152 @@ public static MonitoredResource getResource(String projectId, String resourceTyp
...
-      return Resource.GceInstance;
+  /**
+   * Heuristic to discover the namespace name of the current environment. There is no determenistic
+   * way to discover the namespace name of the process. The name is read from the {@link
+   * K8S_POD_NAMESPACE_PATH} when available or read from a user defined environment variable
+   * "NAMESPACE_NAME"
+   *
+   * @return Namespace name or empty string if the name could not be discovered
+   */
+  private static String getK8sNamespace() {
```
simonz130, line 293: is this method tested with unit test?

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -144,135 +206,152 @@ public static MonitoredResource getResource(String projectId, String resourceTyp
...
+  /**
+   * Distinguish between Standard and Flexible GAE environments. There is no indicator of the
+   * environment. The path to the startup-script in the metadata attribute was selected as one of
+   * the new values that explitly mentioning "flex" and cannot be altered by user (e.g. environment
+   * variable). The method assumes that the resource type is already identified as {@link
+   * Resource.AppEngine}.
+   *
+   * @return "flex" {@link String} for the Flexible environment and "standard" for the Standard.
+   */
+  private static String getAppEngineEnvironment() {
```
simonz130, line 318: is this tested? Maybe worth reaching out to serverless folks (DPE/SWE) to see if this is the best heuristic we can get

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -144,135 +206,152 @@ public static MonitoredResource getResource(String projectId, String resourceTyp
...
 
-  private static String getAppEngineInstanceName() {
-    return System.getenv("GAE_INSTANCE");
+  /**
+   * Retrieves a region from the qualified region of 'projects/[PROJECT_NUMBER]/regions/[REGION]'
+   *
+   * @return region string id
+   */
+  private static String getRegion() {
+    String loc = getter.getAttribute("instance/region");
```
simonz130, line 332: if these strings are used in getter as well, maybe worth extracting them to constants class and using the same string here and in getter

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -144,135 +206,152 @@ public static MonitoredResource getResource(String projectId, String resourceTyp
...
   }
 
   private static List<LoggingEnhancer> createEnhancers(Resource resourceType) {
     List<LoggingEnhancer> enhancers = new ArrayList<>(2);
-    switch (resourceType) {
-        // Trace logging enhancer is supported on GAE Flex and Standard.
-      case GaeAppFlex:
+    if (resourceType == Resource.AppEngine) {
+      enhancers.add(new TraceLoggingEnhancer(APPENGINE_LABEL_PREFIX));
+      if (getAppEngineEnvironment() == FLEX_ENV) {
```
simonz130, line 350: Why the logic is changed here?
We need to handle trace in AppEngine standard as well

simonz130: Also, I didn't see a test that verifies TraceLoggingEnhancer is added

simonz130 (changes requested):
Very solid PR! Thanks!

We need to ensure that TraceEnhancer is treated correctly for AppEngine cases

google-cloud-logging/src/main/java/com/google/cloud/logging/ResourceTypeEnvironmentGetter.java
```diff
@@ -0,0 +1,55 @@
...
+   * @see MetadataConfig#getAttribute()
+   */
+  String getAttribute(String name);
+}
+
+final class ResourceTypeEnvironmentGetterImpl implements ResourceTypeEnvironmentGetter {
+
+  @Override
+  public String getEnv(String name) {
+    return System.getenv(name);
```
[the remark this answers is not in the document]
minherz: Respectfully disagree. The goal of this wrapper is only to allow successful unit testing. Returning `null` is both handled in the code and validated in the tests. Throwing exception can happened in a case of the illegal input which is unit tested.

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
minherz: There are two unit tests that validate discovery of the GAE resource type - one for flexible and another for standard environments.
The info is taken after careful inspection of what environments provide including environment variables, metadata attributes and file system. I choose smth that seemed more or less deterministic. The "legacy" code did not work well. If you like, I will demo you on Monday.

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
minherz: Such a change creates two problems:
1) Overload the getter class with many functions that essentially do the same thing.
2) There are no much use for either of these calls. Almost any attribute is read 1-2 times during the flow, so even caching it does not make much sense.

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
minherz: Because the legacy logic used **two** resource types for GAE and then extra converted them into one. They also populated them with different labels which comes across the public documentation description.
End users still see **single** resource type. So the functionality is not changed but the GAE environment is identified via different condition. I am considering to remove the whole JUL handler logic from this file and review it. 
I think we will need to review the support for JUL and replace it with support of the multiple logging frameworks instead. Logback appender can be used with JUL without losing in functionality.

google-cloud-logging/src/main/java/com/google/cloud/logging/ResourceTypeEnvironmentGetter.java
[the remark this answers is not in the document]
simonz130: The problem is that the code might change hands and evolve. It might be executed in customer environment and throw exceptions causing customer outages. This might be undetected in our tests (e.g. throwing security exception on env variable reading). Our library code needs to be implemented "defensively", so the library is not crashing.

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
simonz130: It can be refactored using visitor pattern, where each node will represent a managed resource with its labels. Example here: https://en.wikipedia.org/wiki/Visitor_pattern#Java_example This way we won't have any "ifs" or "switch" statements and instead have a bunch of classes that represent resources and a method per class that "visits" (detects) the resource. IMO, it's a lot cleaner and easier to test, esp. if we want to expose this logic to other use-cases, such as metrics collection, OpenTelemetry Trace exporters, etc. 

There might be other alternatives available. It's not something we need to do in this PR but I want us to do this "refactoring cleanup" once, so adding more resources will be a lot easier going forward.

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
simonz130: Yes, would love to see a demo. Also worth confirming with @grayside - AppEngine folks work on Titanium release - I want to make sure this logic will continue to work in light of that work

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
simonz130: I didn't mean to create more methods exposing it. I meant creating constants so all strings are in one place:

```java
getter.getAttribute(ResourceStrings.INSTANCE_REGION);
```

```java
class ResourceStrings {
   public const String INSTANCE_REGION = "instance/region";
   ...
   // other similar constants
}
```

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
simonz130: Yeah, I see now that there is only gae_instance (single resource - no matter if it's gae flex or standard). Thanks for clarifying.

simonz130 (approved):

simonz130:
Also note that we have environment tests for Run and Functions failing. Given that this PR adds Functions detection support, worth looking at the env test failure and seeing why is it failing.

google-cloud-logging/src/main/java/com/google/cloud/logging/ResourceTypeEnvironmentGetter.java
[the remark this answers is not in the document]
minherz: Got it, I will add handling of `SecurityException`. `NullPointerException` should not be thrown by our code in any circumstances and if it happens it should be handled by our tests or worst case cause the crash so we can fix it.

minherz:
Function env fails because envctl does not support "functions" environment. @daniel-sanche can you help me with checking the reasons for Cloud Run?

PR #716, opened 2021-10-19
chore: refactor retrieving resource metadata

### LINKED ISSUE #710: structured logging: refactor logic of MonitoredResourceUtil#getValue() to avoid long switch
Current implementation (see #685)  of MonitoredResourceUtil#getValue() is hard to maintain or understand. The [code](https://github.com/googleapis/java-logging/blob/0592c96bc697a983f521c7ac51df53661c801975/google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java#L215-L283) should be refactored, so each label is built in the designated function.
One of possible approaches would be using a [visitor pattern](https://en.wikipedia.org/wiki/Visitor_pattern#Java_example).

Such change will simplify further maintenance (e.g. adding new labels) as well as exposing this functionality for external use.

minherz:
host logic for getting metadata values for monitored resource in MetadataLoader
refactor MonitoredResourceUitl to use MetadataLoader

Fixes #710

### DIFF
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
@@ -20,10 +20,6 @@ import com.google.cloud.MonitoredResource;
 import com.google.cloud.logging.LogEntry.Builder;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableMultimap;
-import java.io.IOException;
-import java.nio.charset.StandardCharsets;
-import java.nio.file.Files;
-import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -37,21 +33,17 @@
 public class MonitoredResourceUtil {
 
   private static final String APPENGINE_LABEL_PREFIX = "appengine.googleapis.com/";
-  private static final String CLUSTER_NAME_ATTRIBUTE = "instance/attributes/cluster-name";
-  private static final String K8S_POD_NAMESPACE_PATH =
-      "/var/run/secrets/kubernetes.io/serviceaccount/namespace";
-  private static final String FLEX_ENV = "flex";
-  private static final String STD_ENV = "standard";
```
simonz130, line 44: do we need this both here and in MetadataLoader? Maybe we can consolidate?

simonz130 (approved):
So much better now! Thanks for this refactoring!

PR #727, opened 2021-10-26
feat: Extend a set of options in WriteOption to allow defining the log location as custom project, folder, organization or billing account

### LINKED ISSUE #684: Extend a set of options in WriteOption to allow defining the log location as custom project, folder, org or billing account
Currently, the set of WriteOption options is limited to 
https://github.com/googleapis/java-logging/blob/9d0e7a28f96ace0189beeb084a747886a6afd55f/google-cloud-logging/src/main/java/com/google/cloud/logging/Logging.java#L69-L72
and does not allow to customize log name. The location of the log is produced based on the project id which is taken from the logging client options:
https://github.com/googleapis/java-logging/blob/9d0e7a28f96ace0189beeb084a747886a6afd55f/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java#L741-L746

This project id is retrieved from authentication workflow and is not necessary the same with the needed log destination. There also cases when the destination should be other than a project

losalex:
Exposed an ability to set logs destination in WriteOption and added an ability to customize log destination in LoggingHandler 

Fixes [#684](https://github.com/googleapis/java-logging/issues/684) ☕️

### DIFF
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
@@ -746,12 +746,6 @@ class LoggingImpl extends BaseService<LoggingOptions> implements Logging {
 
     if (logName != null) {
       builder.setLogName(logName.toString());
-
-      // In case if project ID was provided in request options, it should be used also to
-      // generate log entries in case if those does not have any destination set.
-      if (logName.getProject() != null) {
-        projectId = logName.getProject();
-      }
     }
     MonitoredResource resource = RESOURCE.get(options);
     if (resource != null) {
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
@@ -740,15 +740,18 @@ class LoggingImpl extends BaseService<LoggingOptions> implements Logging {
       Iterable<LogEntry> logEntries,
       Map<Option.OptionType, ?> options) {
     WriteLogEntriesRequest.Builder builder = WriteLogEntriesRequest.newBuilder();
-    String projectId = null;
+    String projectId = serviceOptions.getProjectId();
 
-    LogName logName =
-        getLogName(
-            serviceOptions.getProjectId(), LOG_NAME.get(options), LOG_DESTINATION.get(options));
+    LogName logName = getLogName(projectId, LOG_NAME.get(options), LOG_DESTINATION.get(options));
 
     if (logName != null) {
       builder.setLogName(logName.toString());
-      projectId = logName.getProject();
+
+      // In case if project ID was provided in request options, it should be used also to
+      // generate log entries in case if those does not have any destination set.
+      if (logName.getProject() != null) {
+        projectId = logName.getProject();
+      }
     }
     MonitoredResource resource = RESOURCE.get(options);
     if (resource != null) {
@@ -759,7 +762,6 @@ class LoggingImpl extends BaseService<LoggingOptions> implements Logging {
       builder.putAllLabels(labels);
     }
 
-    if (projectId == null) projectId = serviceOptions.getProjectId();
     builder.addAllEntries(Iterables.transform(logEntries, LogEntry.toPbFunction(projectId)));
     return builder.build();
   }
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
@@ -21,6 +21,7 @@ import static com.google.cloud.logging.Logging.EntryListOption.OptionType.ORDER_
 import static com.google.cloud.logging.Logging.ListOption.OptionType.PAGE_SIZE;
 import static com.google.cloud.logging.Logging.ListOption.OptionType.PAGE_TOKEN;
 import static com.google.cloud.logging.Logging.WriteOption.OptionType.LABELS;
+import static com.google.cloud.logging.Logging.WriteOption.OptionType.LOG_DESTINATION;
 import static com.google.cloud.logging.Logging.WriteOption.OptionType.LOG_NAME;
 import static com.google.cloud.logging.Logging.WriteOption.OptionType.RESOURCE;
 
@@ -738,11 +739,16 @@ class LoggingImpl extends BaseService<LoggingOptions> implements Logging {
       LoggingOptions serviceOptions,
       Iterable<LogEntry> logEntries,
       Map<Option.OptionType, ?> options) {
-    String projectId = serviceOptions.getProjectId();
     WriteLogEntriesRequest.Builder builder = WriteLogEntriesRequest.newBuilder();
-    String logName = LOG_NAME.get(options);
+    String projectId = null;
+
+    LogName logName =
+        getLogName(
+            serviceOptions.getProjectId(), LOG_NAME.get(options), LOG_DESTINATION.get(options));
+
     if (logName != null) {
-      builder.setLogName(LogName.ofProjectLogName(projectId, logName).toString());
+      builder.setLogName(logName.toString());
+      projectId = logName.getProject();
     }
     MonitoredResource resource = RESOURCE.get(options);
     if (resource != null) {
@@ -752,8 +758,24 @@ class LoggingImpl extends BaseService<LoggingOptions> implements Logging {
     if (labels != null) {
       builder.putAllLabels(labels);
     }
+
+    if (projectId == null) projectId = serviceOptions.getProjectId();
     builder.addAllEntries(Iterables.transform(logEntries, LogEntry.toPbFunction(projectId)));
     return builder.build();
+  }
+
+  private static LogName getLogName(
+      String projectId, String logName, LogDestinationName destination) {
+    if (logName == null) {
+      return null;
+    }
+
+    // If no destination specified, fallback to project based log name
+    if (destination == null) {
+      return LogName.ofProjectLogName(projectId, logName);
+    }
+
+    return destination.toLogName(logName);
   }
 
   public void write(Iterable<LogEntry> logEntries, WriteOption... options) {
--- a/google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
+++ b/google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
@@ -127,13 +127,13 @@ public class LoggingImplTest {
   private static final MonitoredResource MONITORED_RESOURCE =
       MonitoredResource.newBuilder("global").addLabel("project-id", PROJECT).build();
   private static final LogEntry LOG_ENTRY1 =
-      LogEntry.newBuilder(StringPayload.of("entry1"))
+      LogEntry.newBuilder(StringPayload.of("entry-1"))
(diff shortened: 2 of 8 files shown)


minherz (changes requested):
We recommend to minimize the scope of changes in order to speed up and simplify the reviewing process. It reduces a cognitive load of the reviewer and helps avoiding mistakes.

[edited] For the next time, please split changes into multiple PRs. In this PR we shall finish reviewing the changes related to the write interface and then will do modifications to `LoggingHandler` and `LoggingHandlerTest`, since the latter changes are not directly related to the PR's description and the linked issue(s).

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
```diff
@@ -738,11 +739,16 @@ private static WriteLogEntriesRequest writeLogEntriesRequest(
...
       Map<Option.OptionType, ?> options) {
-    String projectId = serviceOptions.getProjectId();
     WriteLogEntriesRequest.Builder builder = WriteLogEntriesRequest.newBuilder();
-    String logName = LOG_NAME.get(options);
+    String projectId = null;
+
+    LogName logName =
+        getLogName(
+            serviceOptions.getProjectId(), LOG_NAME.get(options), LOG_DESTINATION.get(options));
+
```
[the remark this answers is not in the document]
minherz: Would you define criteria when a block of code in one method should be branched to a separate method beside the case when this code is reused/called in/from more than one place? It would be great to have a shared guideline for this so we can have a homogeneously looking code.

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
```diff
@@ -1795,25 +1830,32 @@ public void testWriteLogEntriesWithSeverityFlushEnabled() {
...
-    WriteLogEntriesResponse response = WriteLogEntriesResponse.newBuilder().build();
-    EasyMock.expect(loggingRpcMock.write(request)).andReturn(ApiFutures.immediateFuture(response));
-    EasyMock.replay(rpcFactoryMock, loggingRpcMock);
-    logging = options.getService();
-    logging.write(
-        ImmutableList.of(LOG_ENTRY1, LOG_ENTRY2),
-        WriteOption.logName(LOG_NAME),
-        WriteOption.resource(MONITORED_RESOURCE),
-        WriteOption.labels(labels));
+    test_write_log_entries_with_destination(
```
minherz, line 1833: Please fix the method name `test_write_log_entries_with_destination` to follow [Java style](https://google.github.io/styleguide/javaguide.html#s5.2.3-method-names) for method names.

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
```diff
@@ -91,6 +91,10 @@
 public class LoggingImplTest {
 
   private static final String PROJECT = "project";
+  private static final String PROJECT_OVERRIDE = "projectoverride";
```
minherz, line 94: I think that the use of "override" word in the name of the constant introduce a logic that may misguide or mislead a reader. Also it is useful when the name of the constant defines the role of the constant value and not only its content. For example, `PROJECT` constant is used as a project id and will be more readable as `PROJECT_ID` or `DEFAULT_PROJECT_ID`.
I recommend to address all relevant constant names accordingly.

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
```diff
@@ -113,6 +117,12 @@
   private static final String LOG_NAMES_CURSOR = "cursor";
   private static final String LOG_NAME = "log";
   private static final String LOG_NAME_PB = "projects/" + PROJECT + "/logs/" + LOG_NAME;
+  private static final String LOG_NAME_OVERRIDE_PB =
```
minherz, line 120: I am not sure whether you followed existing pattern or not but `_PB` suffix is not informative IMO. The constant value is a fully qualified name but it is not related to protobuf (assuming `_PB` suffix references it).
I recommend to review all constants with this suffix and make their names to be more informative. In this particular case `LOG_NAME` can be renamed to `LOG_ID` and `LOG_NAME_PB` to become `LOG_NAME`.

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
```diff
@@ -1795,25 +1830,32 @@ public void testWriteLogEntriesWithSeverityFlushEnabled() {
...
-        WriteOption.labels(labels));
+    test_write_log_entries_with_destination(
+        PROJECT, LOG_NAME_PB, LogDestinationName.project(PROJECT));
+  }
+
+  @Test
+  public void testWriteLogEntriesWithProjectOverrideOptions() {
+    test_write_log_entries_with_destination(
+        PROJECT, LOG_NAME_OVERRIDE_PB, LogDestinationName.project(PROJECT_OVERRIDE));
+  }
```
minherz, line 1841: what is the difference between this test and the `testWriteLogEntriesWithOptions` test beside different values of the project id in the parameters?
It looks like the same test to me.

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
```diff
@@ -1852,7 +1894,8 @@ public void testWriteLogEntriesAsyncWithOptions() {
         ImmutableList.of(LOG_ENTRY1, LOG_ENTRY2),
         WriteOption.logName(LOG_NAME),
         WriteOption.resource(MONITORED_RESOURCE),
-        WriteOption.labels(labels));
+        WriteOption.labels(labels),
+        WriteOption.destination(LogDestinationName.project(PROJECT)));
```
minherz, line 1898: Do we have a test that validates the behavior of the `write()` **without** destination WriteOption?

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
```diff
@@ -2183,4 +2226,41 @@ public void run() {
...
+            LOG_ENTRY_BILLING,
+            LOG_ENTRY_FOLDER,
+            LOG_ENTRY_ORGANIZATION,
+            LOG_ENTRY2,
+            LOG_ENTRY_EMPTY),
+        WriteOption.logName(LOG_NAME),
+        WriteOption.resource(MONITORED_RESOURCE),
+        WriteOption.labels(labels),
+        WriteOption.destination(destination));
+  }
```
minherz, line 2265: what does this test assert? it looks like there is no comparison between expected and actual results.
i recommend to rewrite the test in a way that it includes validation that given the input the expected results match the actual results.

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingTest.java
```diff
@@ -100,8 +104,33 @@ public void testWriteOption() {
...
+    resource = (LogDestinationName) writeOption.getValue();
+    assertEquals(WriteOption.OptionType.LOG_DESTINATION, writeOption.getOptionType());
+    assertEquals(LogDestinationName.DestinationType.FOLDER, resource.getOptionType());
+    assertEquals(FOLDER_NAME, resource.getValue());
+
+    writeOption = WriteOption.destination(LogDestinationName.organization(ORGANIZATION_NAME));
+    resource = (LogDestinationName) writeOption.getValue();
+    assertEquals(WriteOption.OptionType.LOG_DESTINATION, writeOption.getOptionType());
+    assertEquals(LogDestinationName.DestinationType.ORGANIZATION, resource.getOptionType());
+    assertEquals(ORGANIZATION_NAME, resource.getValue());
```
minherz, line 134: nit: consider breaking it into multiple tests. it will be more visual and easy to modify in future.

minherz (changes requested):
Please, have a look into comments for changes in the LoggingImplTest.java

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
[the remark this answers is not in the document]
losalex: I believe 5-15 lines including parameters is more than enough (so this way entire function can be visible on small laptop screen). Some more good answers you can find [here](https://softwareengineering.stackexchange.com/questions/133404/what-is-the-ideal-length-of-a-method-for-you).

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
losalex: As for LOG_NAME - it is used in calls to setLogName(), so passing LOG_ID might be a bit misleading. I changed some names, please let me know if that OK

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
losalex: It checks cases when project ID used in `LogEntry.toPbFunction()` is same as for request and case when it is different.

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
losalex: Perhaps `testWriteLogEntriesAsync()` answers your question?

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
losalex: It does validates - I believe this happens in `tearDown()`

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
losalex: Renamed override.

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
minherz: I think it is not. I was looking for a test that validating the "previous" logic when a user provides log name option and does not provide destination.

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
minherz: I see. Indeed, the line 2250 should do the trick, but it is very hard to spot.
nit: Would you mind to give to the `request` variable a more "descriptive" name? Something like `expectedWriteLogEntriesRequest` or similar. It will let to a reader to better understand what we assert in the call to `expect()`.

minherz (approved):
LGTM. Feel free to ignore or apply the comment about `request` variable name.

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
losalex: Sure - adding missing part

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
losalex: Sure

losalex:
Thanks!

PR #729, opened 2021-10-28
feat: Implement auto population of the HttpRequest from the current HTTP context

### LINKED ISSUE #687: structured logging: add Http request context discovery
If not defined, the httpRequest field in LogEntry should be auto-populated when writing log entries to backend

Provide a method to define an Http context that can be used by a Web server application to publish info about the request context and also to remove the published context.
The method should support multi-threading. It should be available from any thread of the application process.

minherz:
Implement population of the `HttpRequest` instance using information from the currently available HTTP context.
Implement the HTTP context for the Spring framework web applications.
Add (missing) `protocol` field to the `HttpRequest` class.

Fixes #687


minherz:
discontinue this solution in favor of different approach.

PR #731, opened 2021-11-02
feat: Add an ability to delete logs based on project, folder, organization or billing account resource names

### LINKED ISSUE #725: Add an ability to delete logs based on project, folder, organization or billing account resource names
Recently we introduced an ability to [add a new property to the LogEntry class that describes the resource where the log entry is (or will be) stored](https://github.com/googleapis/java-logging/issues/719). This solution introduces a new problem, when log deletions cannot be done for non-project base log destinations since delete log functionality today uses only project bases resource names.

Add an ability to delete logs based on additional parameters like custom project, folder, organization or billing account resource names.

**Describe alternatives you've considered**
None.

**Additional context**
See more explanation about resource names under [logName field](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry)

losalex:
Add an ability to delete logs based on project, folder, organization or billing account resource names and add test coverage for delete functionality

Fixes [#725](https://github.com/googleapis/java-logging/issues/725) ☕️

### DIFF
--- a/google-cloud-logging/clirr-ignored-differences.xml
+++ b/google-cloud-logging/clirr-ignored-differences.xml
@@ -4,6 +4,6 @@
   <difference>
     <differenceType>7012</differenceType>
     <className>com/google/cloud/logging/Logging</className>
-    <method>* deleteLog*(*)</method>
+    <method>* deleteLog*(java.lang.String, com.google.cloud.logging.LogDestinationName)</method>
   </difference>
 </differences>
--- a/google-cloud-logging/clirr-ignored-differences.xml
+++ b/google-cloud-logging/clirr-ignored-differences.xml
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- see http://www.mojohaus.org/clirr-maven-plugin/examples/ignored-differences.html -->
+<differences>
+  <difference>
+    <differenceType>7012</differenceType>
+    <className>com/google/cloud/logging/Logging</className>
+    <method>* deleteLog*(*)</method>
+  </difference>
+</differences>
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/Logging.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/Logging.java
@@ -602,7 +602,7 @@ public interface Logging extends AutoCloseable, Service<LoggingOptions> {
    * https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry). The log will reappear if
    * new entries are written to it.
    *
-   * <p>Example of deleting a log.
+   * <p>Example of deleting a log by folder destination.
    *
    * <pre>
    * {
@@ -620,7 +620,10 @@ public interface Logging extends AutoCloseable, Service<LoggingOptions> {
    *
    * @return {@code true} if the log was deleted, {@code false} if it was not found
    */
-  boolean deleteLog(String log, LogDestinationName destination);
+  default boolean deleteLog(String log, LogDestinationName destination) {
+    throw new UnsupportedOperationException(
+        "method deleteLog() does not have default implementation");
+  }
 
   /**
    * Sends a request for deleting a log and all its log entries for given log destination (see
@@ -628,7 +631,7 @@ public interface Logging extends AutoCloseable, Service<LoggingOptions> {
    * This method returns a {@code ApiFuture} object to consume the result. {@link ApiFuture#get()}
    * returns {@code true} if the log was deleted, {@code false} if it was not found.
    *
-   * <p>Example of asynchronously deleting a log.
+   * <p>Example of asynchronously deleting a log by folder destination.
    *
    * <pre>
    * {
@@ -646,7 +649,10 @@ public interface Logging extends AutoCloseable, Service<LoggingOptions> {
    * }
    * </pre>
    */
-  ApiFuture<Boolean> deleteLogAsync(String log, LogDestinationName destination);
+  default ApiFuture<Boolean> deleteLogAsync(String log, LogDestinationName destination) {
+    throw new UnsupportedOperationException(
+        "method deleteLogAsync() does not have default implementation");
+  }
 
   /**
    * Sends a request for deleting a log and all its log entries. This method returns a {@code
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/Logging.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/Logging.java
@@ -589,6 +589,57 @@ public interface Logging extends AutoCloseable, Service<LoggingOptions> {
   boolean deleteLog(String log);
 
   /**
+   * Deletes a log and all its log entries for given log destination (see 'logName' parameter in
+   * https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry). The log will reappear if
+   * new entries are written to it.
+   *
+   * <p>Example of deleting a log.
+   *
+   * <pre>
+   * {
+   *   &#64;code
+   *   String logName = "my_log_name";
+   *   String folder = "my_folder";
+   *   boolean deleted = logging.deleteLog(logName, LogDestinationName.folder(folder));
+   *   if (deleted) {
+   *     // the log was deleted
+   *   } else {
+   *     // the log was not found
+   *   }
+   * }
+   * </pre>
+   *
+   * @return {@code true} if the log was deleted, {@code false} if it was not found
+   */
+  boolean deleteLog(String log, LogDestinationName destination);
+
+  /**
+   * Sends a request for deleting a log and all its log entries for given log destination (see
+   * 'logName' parameter in https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry).
+   * This method returns a {@code ApiFuture} object to consume the result. {@link ApiFuture#get()}
+   * returns {@code true} if the log was deleted, {@code false} if it was not found.
+   *
+   * <p>Example of asynchronously deleting a log.
+   *
+   * <pre>
+   * {
+   *   &#64;code
+   *   String logName = "my_log_name";
+   *   String folder = "my_folder";
+   *   ApiFuture<Boolean> future = logging.deleteLogAsync(logName, LogDestinationName.folder(folder));
+   *   // ...
+   *   boolean deleted = future.get();
+   *   if (deleted) {
+   *     // the log was deleted
+   *   } else {
+   *     // the log was not found
+   *   }
+   * }
+   * </pre>
+   */
+  ApiFuture<Boolean> deleteLogAsync(String log, LogDestinationName destination);
+
+  /**
    * Sends a request for deleting a log and all its log entries. This method returns a {@code
    * ApiFuture} object to consume the result. {@link ApiFuture#get()} returns {@code true} if the
    * log was deleted, {@code false} if it was not found.
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
@@ -39,6 +39,7 @@ import com.google.cloud.PageImpl;
 import com.google.cloud.logging.spi.v2.LoggingRpc;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Function;
+import com.google.common.base.Preconditions;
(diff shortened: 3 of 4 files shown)


google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
```diff
@@ -443,11 +444,23 @@ public boolean deleteLog(String log) {
...
+    return deleteLogAsyncImpl(log, null);
+  }
+
+  public ApiFuture<Boolean> deleteLogAsync(String log, LogDestinationName destination) {
+    return deleteLogAsyncImpl(log, destination);
+  }
+
+  private ApiFuture<Boolean> deleteLogAsyncImpl(String log, LogDestinationName destination) {
+    LogName name =
+        Preconditions.checkNotNull(getLogName(getOptions().getProjectId(), log, destination));
```
simonz130, line 461: can getOptions() or getOptions().getProjectId() also return null?

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
```diff
@@ -1777,6 +1777,50 @@ public void testDeleteLogAsync() throws ExecutionException, InterruptedException
     assertTrue(logging.deleteLogAsync(LOG_NAME).get());
   }
 
+  @Test
+  public void testDeleteLogBillingDestination() throws ExecutionException, InterruptedException {
+    testDeleteByBestination(
```
simonz130, line 1782: typo: Destination

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
```diff
@@ -2241,6 +2285,20 @@ public void run() {
     assertSame(0, exceptions.get());
   }
 
+  private void testDeleteByBestination(
```
simonz130, line 2288: Destination

simonz130 (approved):

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
losalex: `getOptions()` might return null since `LoggingImpl` constructor accepts null. Thats said is it safe to assume that `getOptions()` is not null at this point since options are accessed already in `LoggingImpl` constructor. If `getProjectId()` is null it will throw an exception during `LogName` building. I will change a code to give meaningful message for caller since for `DeleteLogRequest` the `LogName` construction is mandatory

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
losalex: Oh, right - thanks!

google-cloud-logging/src/main/java/com/google/cloud/logging/Logging.java
```diff
@@ -597,6 +597,57 @@ public static TailOption project(String project) {
...
+   * https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry). The log will reappear if
+   * new entries are written to it.
+   *
+   * <p>Example of deleting a log.
+   *
+   * <pre>
+   * {
+   *   &#64;code
+   *   String logName = "my_log_name";
+   *   String folder = "my_folder";
```
minherz, line 611: nit: worth to emphasis these are folder *ID* and log *ID* to compose the qualified log name.

google-cloud-logging/clirr-ignored-differences.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- see http://www.mojohaus.org/clirr-maven-plugin/examples/ignored-differences.html -->
+<differences>
+  <difference>
+    <differenceType>7012</differenceType>
+    <className>com/google/cloud/logging/Logging</className>
+    <method>* deleteLog*(*)</method>
```
minherz, line 7: nit: please, narrow the declaration at least the arguments. It will emphasis which methods were actually added since the last major version upgrade.

google-cloud-logging/src/main/java/com/google/cloud/logging/Logging.java
```diff
@@ -597,6 +597,57 @@ public static TailOption project(String project) {
...
+   *     // the log was deleted
+   *   } else {
+   *     // the log was not found
+   *   }
+   * }
+   * </pre>
+   *
+   * @return {@code true} if the log was deleted, {@code false} if it was not found
+   */
+  boolean deleteLog(String log, LogDestinationName destination);
```
minherz, line 623: nit: consider providing default definition. this allows to handle a case when users derive from the interface in some way but then upgrade to the next minor version of the library.

google-cloud-logging/src/main/java/com/google/cloud/logging/Logging.java
```diff
@@ -597,6 +597,57 @@ public static TailOption project(String project) {
...
+   *   boolean deleted = future.get();
+   *   if (deleted) {
+   *     // the log was deleted
+   *   } else {
+   *     // the log was not found
+   *   }
+   * }
+   * </pre>
+   */
+  ApiFuture<Boolean> deleteLogAsync(String log, LogDestinationName destination);
```
minherz, line 649: nit: consider providing default definition. this allows to handle a case when users derive from the interface in some way but then upgrade to the next minor version of the library.

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
```diff
@@ -443,11 +444,21 @@ public boolean deleteLog(String log) {
...
+  public boolean deleteLog(String log, LogDestinationName destination) {
+    return get(deleteLogAsyncImpl(log, destination));
+  }
+
   public ApiFuture<Boolean> deleteLogAsync(String log) {
-    DeleteLogRequest request =
-        DeleteLogRequest.newBuilder()
-            .setLogName(LogName.ofProjectLogName(getOptions().getProjectId(), log).toString())
-            .build();
+    return deleteLogAsyncImpl(log, null);
```
minherz, line 452: nit: why `deleteLogAsyncImpl` is needed? why not to call `deleteLogAsync`?

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
```diff
@@ -443,11 +444,21 @@ public boolean deleteLog(String log) {
...
+  }
+
+  public ApiFuture<Boolean> deleteLogAsync(String log, LogDestinationName destination) {
+    return deleteLogAsyncImpl(log, destination);
```
minherz, line 456: nit: why extra call to `deleteLogAsyncImpl` is needed? why not to keep implementation in `deleteLogAsync`?

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
```diff
@@ -443,11 +444,21 @@ public boolean deleteLog(String log) {
...
+  }
+
+  private ApiFuture<Boolean> deleteLogAsyncImpl(String log, LogDestinationName destination) {
+    LogName name = getLogName(getOptions().getProjectId(), log, destination, true);
```
minherz, line 460: nit: mind that passing `true` you modify the existing behavior of the `deleteLog*(String)` method. I am OK with either option: throwing NPE or failing at backend, but it should be unanimous for all methods (not only `deleteLog*`).

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
```diff
@@ -742,7 +753,8 @@ private static WriteLogEntriesRequest writeLogEntriesRequest(
     WriteLogEntriesRequest.Builder builder = WriteLogEntriesRequest.newBuilder();
     String projectId = serviceOptions.getProjectId();
 
-    LogName logName = getLogName(projectId, LOG_NAME.get(options), LOG_DESTINATION.get(options));
+    LogName logName =
+        getLogName(projectId, LOG_NAME.get(options), LOG_DESTINATION.get(options), false);
```
minherz, line 757: see comment to line 460

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
```diff
@@ -761,13 +773,21 @@ private static WriteLogEntriesRequest writeLogEntriesRequest(
...
   private static LogName getLogName(
-      String projectId, String logName, LogDestinationName destination) {
-    if (logName == null) {
+      String projectId,
+      String logName,
+      LogDestinationName destination,
+      boolean validateParameters) {
+    if (validateParameters) {
+      Preconditions.checkNotNull(logName, "logName parameter cannot be null");
+    } else if (logName == null) {
```
minherz, line 782: it is not best solution. more visual and easy to understand would be to validate for NPE before you call this (private) method and keep previous implementation or _always_ throw NPE if required argument(s) are `null`.

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
```diff
@@ -761,13 +773,21 @@ private static WriteLogEntriesRequest writeLogEntriesRequest(
...
       return null;
     }
 
     // If no destination specified, fallback to project based log name
     if (destination == null) {
+      if (validateParameters) {
+        Preconditions.checkNotNull(projectId, "projectId parameter cannot be null");
+      }
```
minherz, line 790: same here

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
```diff
@@ -1777,6 +1777,50 @@ public void testDeleteLogAsync() throws ExecutionException, InterruptedException
     assertTrue(logging.deleteLogAsync(LOG_NAME).get());
   }
 
+  @Test
+  public void testDeleteLogBillingDestination() throws ExecutionException, InterruptedException {
```
minherz, line 1781: nit: there is a possible inconsistency how tests validate different destinations in different interfaces. there are tests which do everything in one method: validating folder, project, org and billing acc destinations and like this change do it separately. it would be great to decide on one approach and follow it in all our tests.

google-cloud-logging/src/main/java/com/google/cloud/logging/Logging.java
losalex: Oh, right - it somehow missed during merge

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
losalex: Yep, we can delete deleteLogAsyncImpl

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
losalex: I prefer to reuse code even in tests

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
losalex: Nope, `LogName` constructor is using `Preconditions.checkNotNull` and previous code was calling the `LogName.ofProjectLogName` which constructs `LogName`

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
losalex: Same as I mentioned above

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
losalex: I agree that in this case we can move parameters validation into `deleteLogAsync` - we can always move those validation into `getLogName()` in case if it will be called from more places in the future (since otherwise we are duplicating checks all over the code).

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
losalex: Answered above

PR #759, opened 2021-11-24
fix: handle null pointer when parsing metadata attributes

minherz:
Fix behavior of getRegion and getZone methods when attributes aren't defined.
Align getNamespaceName return value with other methods to return null.

Addresses googleapis/java-logging-logback#599 when incorrect resource type is enforced.

### DIFF
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/MetadataLoader.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/MetadataLoader.java
@@ -87,8 +87,8 @@ public final class MetadataLoader {
    * Distinguish between Standard and Flexible GAE environments. There is no indicator of the
    * environment. The path to the startup-script in the metadata attribute was selected as one of
    * the new values that explitly mentioning "flex" and cannot be altered by user (e.g. environment
-   * variable). The method assumes that the resource type is already identified as {@link
-   * Resource.AppEngine}.
+   * variable).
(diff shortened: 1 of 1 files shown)


google-cloud-logging/src/main/java/com/google/cloud/logging/MetadataLoader.java
```diff
@@ -178,7 +175,10 @@ private String getProjectId() {
    */
   private String getRegion() {
     String loc = getter.getAttribute("instance/region");
-    return loc.substring(loc.lastIndexOf('/') + 1);
+    if (loc != null) {
```
losalex, line 178: nit: return loc != null ? loc.substring(loc.lastIndexOf('/') + 1) : null;

google-cloud-logging/src/main/java/com/google/cloud/logging/MetadataLoader.java
```diff
@@ -199,6 +199,9 @@ private String getVersionId() {
    */
   private String getZone() {
     String loc = getter.getAttribute("instance/zone");
-    return loc.substring(loc.lastIndexOf('/') + 1);
+    if (loc != null) {
+      return loc.substring(loc.lastIndexOf('/') + 1);
```
losalex, line 203: nit: Same as above - simplifies reading

losalex (approved):
LGTM

PR #768, opened 2021-12-01
chore(java7,clirr): make clirr to compare vs 2.x

minherz:
by default clirr plugin compares vs. latest (for maven) release of the package which is (for logging) 3.x.
java7 branch builds 2.x versions (current 2.4.0). this change customizes clirr configuration for logging
to make clirr to compare vs. previous release (2.3.0). there is no way to automate this process i.e. update 2.3.0 to 2.4.0 when java7 branch rolls to next minor version. this change will have to be done manually.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -138,30 +138,12 @@
           <artifactId>clirr-maven-plugin</artifactId>
           <version>2.8</version>
           <configuration>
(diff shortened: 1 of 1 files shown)


losalex (approved):
LGTM

PR #777, opened 2021-12-03
fix: enforce w3c trace context value validation

### LINKED ISSUE #774: Fix validation of the W3C trace context header parsing
Parsing should validate the length of each field and allow only hexadecimal characters in the values.

minherz:
Validates length and characters of the w3c trace context value. See [W3C standard](https://www.w3.org/TR/trace-context/#traceparent-header-field-values) for more details.
Refactor unit testing to bring invalid w3c trace context tests to separate file.

Fixes #774


google-cloud-logging/src/main/java/com/google/cloud/logging/Context.java
```diff
@@ -18,11 +18,17 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /** Class to hold context attributes including information about {@see HttpRequest} and tracing. */
 public class Context {
+  // validate W3C trace context value on load according to the existing version format.
+  // see https://www.w3.org/TR/trace-context/#traceparent-header-field-values for details.
+  private static final Pattern W3C_TRACE_CONTEXT_FORMAT =
+      Pattern.compile(
+          "^00-(?!00000000000000000000000000000000)[0-9abcdef]{32}-(?!000000000000)[0-9abcdef]{12}-[0-9abcdef]{2}$");
```
daniel-sanche, line 31: I think the span field is supposed to be 16 characters (see [example here](https://www.w3.org/TR/trace-context/#relationship-between-the-headers))

FYI, the nodejs library uses [slightly different regex](https://github.com/googleapis/nodejs-logging/blob/005cced4f8c1fe5f90496537e4731c2272946cb7/src/context.ts#L220), which I [copied in Python](https://github.com/googleapis/python-logging/pull/450). There may be value in trying to use the same expressions across libraries

google-cloud-logging/src/main/java/com/google/cloud/logging/Context.java
```diff
@@ -142,26 +148,15 @@ public Builder loadCloudTraceContext(String cloudTrace) {
...
     public Builder loadW3CTraceParentContext(String traceParent) throws IllegalArgumentException {
       if (traceParent != null) {
-        String[] fields = traceParent.split("-");
-        if (fields.length > 3) {
-          String versionFormat = fields[0];
-          if (!versionFormat.equals("00")) {
-            throw new IllegalArgumentException("Not supporting versionFormat other than \"00\"");
-          }
-        } else {
+        Matcher validator = W3C_TRACE_CONTEXT_FORMAT.matcher(traceParent.toLowerCase());
```
daniel-sanche, line 151: I'm not sure if `toLowerCase()` is needed. The [spec says](https://www.w3.org/TR/trace-context/#parent-id) "Vendors MUST ignore the traceparent when the parent-id is invalid (for example, if it contains non-lowercase hex characters)."

google-cloud-logging/src/main/java/com/google/cloud/logging/Context.java
minherz: And you would be absolutely correct. Late hours are never good for commits :smile:

google-cloud-logging/src/main/java/com/google/cloud/logging/Context.java
minherz: I do not know whether we should be so strict about it. A chance that two different trace ids will be produced that differ only by a specific letter register seems low to me. But this way we can avoid exceptions that interrupt the metadata collection in the library.

losalex: Should we use then toUpperCase() to be safe?

daniel-sanche: My vote would be to just ignore any `traceparent` headers with invalid characters as the spec suggests, that upper-case "MUST" in the docs seems pretty serious haha. 

My interpretation is that header parsing is meant to be farily strict, we're not supposed to alter the fields to make the data fit our expectations. But I'll leave the decision up to you.

daniel-sanche (approved):

google-cloud-logging/src/main/java/com/google/cloud/logging/Context.java
minherz: I agree with you about the standard. My concern is that ignoring it (or throwing and catching exception on the same account) would be very hard to troubleshoot for application developers. To be a little more liberal on this might prove to be closer to the spirit of the shared library.

PR #796, opened 2021-12-10
fix: Expose ID and destination type of LogDestinationName publicly

losalex:
Expose ID and destination type of LogDestinationName publicly
This is additional fix for [#684](https://github.com/googleapis/java-logging/issues/684) ☕️

### DIFF
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/LogDestinationName.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/LogDestinationName.java
@@ -29,7 +29,7 @@ import java.util.Map;
 public final class LogDestinationName extends Option {
   private static final long serialVersionUID = 7944256748441111191L;
 
-  enum DestinationType implements Option.OptionType {
+  public enum DestinationType implements Option.OptionType {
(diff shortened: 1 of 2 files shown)


google-cloud-logging/src/main/java/com/google/cloud/logging/LogDestinationName.java
```diff
@@ -109,6 +109,16 @@ public LogName toLogName(String logId) {
     return null;
   }
 
+  /** Returns ID value associated with {@code LogDestinationName} object */
+  public String getId() {
+    return getValue().toString();
```
daniel-sanche, line 114: nit: calling this an "id" is a bit confusing, since you also have `toLogName(logId)`

Maybe this should be called `getDestinationId()`?

daniel-sanche (approved):

google-cloud-logging/src/main/java/com/google/cloud/logging/LogDestinationName.java
losalex: Yep, makes sense

PR #801, opened 2021-12-16
Add requestsize configuration into auto-label.yaml

losalex:
Enable pull request size bot configuration in auto-label.yaml


minherz (approved):
lgtm

PR #807, opened 2021-12-23
feat: configure opt-out metada auto-population for JUL handler

minherz:
Allow to opt-out auto-population of the log entries with additional metadata in JUL handler using the flag in `LoggingOptions` and `LoggingConfig` configuration.

Opt-out is set if at least one of the above will be set to `false`. Absence of the configuration is interpreted same as setting the configuration to `true`.

### DIFF
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
@@ -382,8 +382,6 @@ public class LoggingHandler extends Handler {
 
   /**
    * Sets the metadata auto population flag.
-   *
-   * @param synchronicity {@link Synchronicity}
    */
   public void setAutoPopulateMetadata(Boolean value) {
     checkNotNull(value);
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
@@ -17,6 +17,7 @@
 package com.google.cloud.logging;
 
 import static com.google.common.base.MoreObjects.firstNonNull;
+import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.cloud.MonitoredResource;
 import com.google.cloud.logging.Logging.WriteOption;
@@ -25,6 +26,7 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Iterables;
 import java.time.Instant;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -417,4 +456,8 @@ private Logging getLogging() {
     }
     return logging;
   }
+
+  private static boolean isTrueOrNull(Boolean b) {
```
losalex, line 460: nit: Such methods ideally should accept additional variable with default value to be returned in case if checked parameter is null.

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -365,6 +380,30 @@ public Synchronicity getSynchronicity() {
     return getLogging().getWriteSynchronicity();
   }
 
+  /**
+   * Sets the metadata auto population flag.
+   *
+   * @param synchronicity {@link Synchronicity}
```
losalex, line 386: Seems param comment is not relevant?

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -365,6 +380,30 @@ public Synchronicity getSynchronicity() {
...
+   */
+  public void setAutoPopulateMetadata(Boolean value) {
```
losalex, line 388: Do you need to make `setAutoPopulateMetadata` to be a thread safe?

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
minherz: The method is called to answer the question whether the parameter is `null` or `Boolean.TRUE`.

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
minherz: OMG. Will fix this one :+1:

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
losalex: I was kinda suggesting to make it "getter", so it can be reused later if needed

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
minherz: While it is possible that the same handler will be reused from multiple threads. It is less probable that multiple configuration adjustments will be run for the same handler from multiple thread in parallel.
Looking on the rest of the configuration "setters", I decided that `volatile` modifier is sufficient. WDYT?

losalex: OK, lets keep it this way

losalex (approved):

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
minherz: There is "getter": https://github.com/googleapis/java-logging/blob/6bc345c6e9ad0d4b3ea5dc51daf8274d8ee09c3e/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java#L401-L403

The above is a service method used at initialization only to enforce opt-out if one of the original sources of the configuration: LoggingOptions or config file has it set to FALSE. However, if the configuration is omitted then it should be interpreted as "allowed" according to the design requirement: to allow by default.

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
minherz: Done. (lines above are marked with `Outdated`)

PR #812, opened 2021-12-30
feat: ingest structured logs using stdout in LoggingHandler

minherz:
Add configuration (edited:) `redirectoToStdout`~~`useStructuredLogging`~~ to `LoggingHandler` class to support an option to print logs to STDOUT formatted as a one line Json and following the [structured logging](https://cloud.google.com/logging/docs/structured-logging) guidelines.
Implement Jsonification of the `LogEntry` object following the structured logging format.
Move metadata population to a standalone API to be used by `LoggingHandler` explicitly.

### DIFF
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
@@ -19,9 +19,17 @@ package com.google.cloud.logging;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.cloud.MonitoredResource;
+import com.google.cloud.logging.Payload.Type;
 import com.google.common.base.Function;
 import com.google.common.base.MoreObjects;
 import com.google.common.collect.ImmutableMap;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonPrimitive;
+import com.google.gson.JsonSerializationContext;
+import com.google.gson.JsonSerializer;
 import com.google.logging.v2.LogEntryOperation;
 import com.google.logging.v2.LogEntrySourceLocation;
 import com.google.logging.v2.LogName;
@@ -61,8 +69,8 @@ public class LogEntry implements Serializable {
   private final HttpRequest httpRequest;
   private final Map<String, String> labels;
   private final Operation operation;
-  private final Object trace;
-  private final Object spanId;
+  private final String trace;
+  private final String spanId;
   private final boolean traceSampled;
   private final SourceLocation sourceLocation;
   private final Payload<?> payload;
@@ -80,8 +88,8 @@ public class LogEntry implements Serializable {
     private HttpRequest httpRequest;
     private Map<String, String> labels = new HashMap<>();
     private Operation operation;
-    private Object trace;
-    private Object spanId;
+    private String trace;
+    private String spanId;
     private boolean traceSampled;
     private SourceLocation sourceLocation;
     private Payload<?> payload;
@@ -245,7 +253,7 @@ public class LogEntry implements Serializable {
      * relative resource name, the name is assumed to be relative to `//tracing.googleapis.com`.
      */
     public Builder setTrace(Object trace) {
-      this.trace = trace;
+      this.trace = trace != null ? trace.toString() : null;
       return this;
     }
 
@@ -257,7 +265,7 @@ public class LogEntry implements Serializable {
 
     /** Sets the ID of the trace span associated with the log entry, if any. */
     public Builder setSpanId(Object spanId) {
-      this.spanId = spanId;
+      this.spanId = spanId != null ? spanId.toString() : null;
       return this;
     }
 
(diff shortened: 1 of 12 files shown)


google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
```diff
@@ -575,6 +583,145 @@ public Builder toBuilder() {
...
+        .appendField("logging.googleapis.com/sourceLocation", sourceLocation)
+        .appendField("logging.googleapis.com/spanId", spanId)
+        .appendField("logging.googleapis.com/trace", trace)
+        .appendField("logging.googleapis.com/trace_sampled", traceSampled);
+    if (payload.getType() == Type.STRING) {
+      formatter.appendField("message", payload.getData(), true);
+    } else if (payload.getType() == Type.JSON) {
+      Payload.JsonPayload jsonPayload = (Payload.JsonPayload) payload;
+      formatter.appendJson(jsonPayload.getDataAsMap(), true);
+    } else if (payload.getType() == Type.PROTO) {
```
losalex, line 718: Would it be more right approach to validate `payload.getType() == Type.PROTO` when payload is set? This way it could throw exception on init stage which IMO more correct approach. Also worth mentioning that supported types needs to be documented (e.g. what is supported and what not).

google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
```diff
@@ -575,6 +583,145 @@ public Builder toBuilder() {
...
+          builder.append(json.substring(0, json.length() - 1).substring(1));
+          if (!appendComma) {
+            builder.append(",");
+          }
+        }
+      }
+      return this;
+    }
+
+    public StructuredLogFormatter appendJson(Object value) {
```
losalex, line 689: Is this method used anywhere? If not, perhaps we can omit it completely?

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -372,34 +401,36 @@ public void setFlushLevel(Level flushLevel) {
...
   /** Gets the metadata auto population flag. */
   public Boolean getAutoPopulateMetadata() {
     return this.autoPopulateMetadata;
   }
 
+  /**
+   * Enable/disable use of structured logging. When enabled, logs will be printed to STDOUT in the
+   * structured logging format. Otherwise, logs will be ingested using Logging API.
+   */
+  public void setUseStructuredLogging(boolean value) {
```
losalex, line 426: Reading the comment, `setUseStructuredLogging` name is misleading - in both cases we use structured logging but when `setUseStructuredLogging` is true we just printing to STDOUT. I would change it to something like `setPrintStructuredLoggingToStdout` or so

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
```diff
@@ -792,6 +792,50 @@ private static LogName getLogName(
     return destination.toLogName(logName);
   }
 
+  public Iterable<LogEntry> populateMetadata(
+      Iterable<LogEntry> logEntries,
+      MonitoredResource customResource,
+      String... exclusionClassPaths) {
+    final Boolean needDebugInfo =
```
losalex, line 799: Do we need to validate `logEntries` for null first?

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
```diff
@@ -792,6 +792,50 @@ private static LogName getLogName(
...
+    final MonitoredResource resourceMetadata =
+        customResource == null
+            ? MonitoredResourceUtil.getResource(getOptions().getProjectId(), null)
+            : customResource;
+    final Context context = (new ContextHandler()).getCurrentContext();
+    final ArrayList<LogEntry> populatedLogEntries = Lists.newArrayList();
+
+    // populate empty metadata fields of log entries before calling write API
+    for (LogEntry entry : logEntries) {
+      LogEntry.Builder builder = entry.toBuilder();
```
losalex, line 816: Can `entry` be null here?

losalex, line 816: Can you rename it to `entryBuilder`? Later it uses validations with `entry` object (e.g. `entry.getResource() == null`), and it a bit misleading later that we check `entry` while operating on a `builder`...

losalex (commented):
Thanks for your effort - good job! I left some comments, PTAL and let me know if those makes sense

google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
```diff
@@ -575,6 +583,145 @@ public Builder toBuilder() {
...
 
+  /**
+   * Customized serializers to match the expected format for timestamp, source location and request
+   * method
+   */
+  static final class InstantSerializer implements JsonSerializer<Instant> {
+    @Override
+    public JsonElement serialize(
+        Instant src, java.lang.reflect.Type typeOfSrc, JsonSerializationContext context) {
+      return new JsonPrimitive(src.toString());
```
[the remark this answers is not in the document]
minherz: This gson "contract" guarantees the expected behavior. I think that extra checks are not necessary because it would mean incorrect operation of the gson which is a bigger problem.

google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
```diff
@@ -575,6 +583,145 @@ public Builder toBuilder() {
...
+    }
+  }
+
+  static final class SourceLocationSerializer implements JsonSerializer<SourceLocation> {
+    @Override
+    public JsonElement serialize(
+        SourceLocation src, java.lang.reflect.Type typeOfSrc, JsonSerializationContext context) {
+      JsonObject obj = new JsonObject();
+      if (src.getFile() != null) {
```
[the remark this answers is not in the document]
minherz: This is the same gson "contract" in all serializers.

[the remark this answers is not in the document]
minherz: Yes, the `SourceLocation` class is a "simple" data class with getters.

google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
```diff
@@ -575,6 +583,145 @@ public Builder toBuilder() {
...
+    }
+  }
+
+  static final class RequestMethodSerializer implements JsonSerializer<HttpRequest.RequestMethod> {
+    @Override
+    public JsonElement serialize(
+        HttpRequest.RequestMethod src,
+        java.lang.reflect.Type typeOfSrc,
+        JsonSerializationContext context) {
+      return new JsonPrimitive(src.name());
```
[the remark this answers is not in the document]
minherz: This is the same gson "contract" in all serializers.

google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
```diff
@@ -575,6 +583,145 @@ public Builder toBuilder() {
...
+          new GsonBuilder()
+              .registerTypeAdapter(Instant.class, new InstantSerializer())
+              .registerTypeAdapter(SourceLocation.class, new SourceLocationSerializer())
+              .registerTypeAdapter(HttpRequest.RequestMethod.class, new RequestMethodSerializer())
+              .create();
+      this.builder = builder;
+    }
+
+    /**
+     * Adds a Json field and value pair to the current string representation. Method does not
```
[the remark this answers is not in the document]
minherz: Please, elaborate what do you mean by "Comment incomplete".
The class is the private nested class and is only used with explicitly set constants. Its misuse should fail as soon as possible. It is expected to fail as soon as possible.

google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
```diff
@@ -575,6 +583,145 @@ public Builder toBuilder() {
...
+      }
+      return this;
+    }
+
+    public StructuredLogFormatter appendField(String name, Object value) {
+      return appendField(name, value, false);
+    }
+
+    /**
+     * Adds a collection of Json fields that {@code value} parameter is serialized to to the current
```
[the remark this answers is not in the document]
minherz: I think that the method can be refactored by explicitly accept a dictionary instead of the base `Object` class. This way the comment can be changed to smth like:
`Serializes a collection of key, value pairs from the {@code value} dictionary to the Json representation.`
Also, the name of the method might worth to be changed to `appendDict`.
WDYT?

google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
minherz: Agree about throwing exception earlier although the code reads more clear this way IMO. The method is expected to be used mainly via handlers (JUL or Logback) which do not produce log entries with protoPayload.
I will submit a change tothe [public documentation](https://cloud.google.com/logging/docs/structured-logging) to add a note about this fact.

google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
minherz: Thank you for spotting it. I will remove this leftover!. Strange that IDE did not highlight it...

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
minherz: Good catch. Indeed, our internal use of the term does not exactly correlate with what public documentation says:
> structured logs refer to log entries that use the jsonPayload field to add structure to their payloads.

The flag name will be changed to `useLoggingAgent`. WDYT?

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
minherz: Will do.

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
minherz: In theory it can. I will add a check, thanx.

minherz: Will do.

google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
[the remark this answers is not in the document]
losalex: Comment does not mention that it will throw exception if `name` is null. However. you do mention in comment that nothing is added when `value` is null - I believe you should add comment for `name` cannot be null

google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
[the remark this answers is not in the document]
losalex: Yes, I think name change like `appendDictionary` will be fine then

google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
losalex: Thank you!

losalex (approved):
Overall looks good - just fix whatever we agreed in comments. Also left some clarification regarding incomplete comment, hope it make sense

google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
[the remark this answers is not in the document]
minherz: The javadoc of the parameter `name` describes it as a valid Json name:
https://github.com/googleapis/java-logging/blob/4c0561c37955c3b97f1f79e44c77a6c44e28b569/google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java#L647

PR #820, opened 2022-01-06
chore: update README about using metadata auto-population and redirecting JUL handler to STDOUT

minherz:
Add description into README about using automatic metadata population and redirect the JUL handlerto stdout using structured logging.
Refactor code to make it more readable.


minherz:
Merging into wrong branch.

PR #880, opened 2022-02-24
docs(sample): Add sample for Native Image support in Logging 

mpeddada1:
This PR copies over the Logging sample, Setup Instructions and README for native image support from [GoogleCloudPlatform/native-image-support-java](https://github.com/GoogleCloudPlatform/native-image-support-java/tree/master/native-image-samples/native-image-samples-client-library/logging-sample). It  also adds an integration test that can be run as a native image.

Calling `mvn package -Pnative -DskipTests` builds the native image executable for the application and calling `mvn test -Pnative` runs the test as a native image. 

For more information: https://graalvm.github.io/native-build-tools/latest/maven-plugin.html#configuration


samples/native-image-sample/src/main/java/com/example/logging/NativeImageLoggingSample.java
```diff
@@ -0,0 +1,63 @@
+/*
+ * Copyright 2020-2021 Google LLC
```
losalex, line 2: Should copyright dates be updated?

samples/native-image-sample/src/test/java/com/example/logging/ITNativeImageLoggingSample.java
```diff
@@ -0,0 +1,45 @@
...
+package com.example.logging;
+
+import static com.google.common.truth.Truth.assertThat;
+
+import java.io.ByteArrayOutputStream;
+import java.io.PrintStream;
+import org.junit.Before;
+import org.junit.Test;
+
+public class ITNativeImageLoggingSample {
```
suztomo, line 26: This test is not in "Kokoro Samples" build.

https://source.cloud.google.com/results/invocations/26ead732-1ea4-4779-aa2e-1a1a6a09ab14/targets

mpeddada1: Yeah I was noticing this too. Hm not sure why it's not added to the Kokoro Samples build. Will have to look into this further.

suztomo: But the source-code comment of build.sh says the build runs if it's not associated with a pull request.

mpeddada1: You're right: https://github.com/googleapis/java-logging/blob/d2a125aa3bf91cfcd01e45d91fa116d645230dca/.kokoro/build.sh#L79. Looks like only ITs in snapshot/ are run in the presubmit build.

samples/native-image-sample/pom.xml
```diff
@@ -0,0 +1,169 @@
...
+  -->
+  <parent>
+    <groupId>com.google.cloud.samples</groupId>
+    <artifactId>shared-configuration</artifactId>
+    <version>1.2.0</version>
+  </parent>
+
+  <properties>
+    <maven.compiler.target>11</maven.compiler.target>
+    <maven.compiler.source>11</maven.compiler.source>
```
suztomo, line 22: This will fail.

suztomo (approved):
Approving but the Java version should be 1.8.

losalex:
> Approving but the Java version should be 1.8.

Is Java version change considered as major upgrade?

samples/native-image-sample/pom.xml
mpeddada1: Good catch.

mpeddada1:
> Is Java version change considered as major upgrade?

This java version change downgrades Java 11 to Java 8. We had to do this because using Java 11 causes the Kokoro Samples nightly job to fail with the following error:

losalex:
> > Is Java version change considered as major upgrade?
> 
> This java version change downgrades Java 11 to Java 8. We had to do this because using Java 11 causes the Kokoro Samples nightly job to fail with the following error:
> 
> 

I am not sure this is safe move considering that Java downgrade might impact customers using this library. Perhaps we can talk about it to get full understanding of the fix?

mpeddada1:
> I am not sure this is safe move considering that Java downgrade might impact customers using this library. Perhaps we can talk about it to get full understanding of the fix?

Oops sorry, didn't mean to edit your comment. For sure! we can discuss this further offline as well. Also adding reasoning for the change here to provide some context: It looks like all the samples in the repo use [Java 8 for compilation](https://github.com/googleapis/java-logging/blob/dc0f893617d05345d1cbbf13c257d1fca389e599/samples/snippets/pom.xml#L21) since the Kokoro nightly jobs use [that version](https://github.com/googleapis/java-logging/blob/dc0f893617d05345d1cbbf13c257d1fca389e599/.kokoro/nightly/samples.cfg#L6) to run the integration tests for the samples. It looks like this is the only option at the moment until we upgrade the Java version on Kokoro. Open to hearing your thoughts on this.

losalex (approved):
Thanks!

PR #979, opened 2022-06-21
feat: Add support for library instrumentation

losalex:
This feature provides an ability to log extra entry with diagnostics structure which contains logging library information. Such entry is logged only once when first entry is written by a process using logging library.

Fixes #[978](https://github.com/googleapis/java-logging/issues/978), [751](https://github.com/googleapis/java-logging/issues/751) 🦕

### DIFF
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/Instrumentation.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/Instrumentation.java
@@ -38,6 +38,7 @@ public class Instrumentation {
   public static final String INSTRUMENTATION_VERSION_KEY = "version";
   public static final String JAVA_LIBRARY_NAME_PREFIX = "java";
   public static final String DEFAULT_INSTRUMENTATION_VERSION = "UNKNOWN";
+  public static final String INSTRUMENTATION_LOG_NAME = "diagnostic-log";
   public static final int MAX_DIAGNOSTIC_VALUE_LENGTH = 14;
   public static final int MAX_DIAGNOSTIC_ENTIES = 3;
   private static boolean instrumentationAdded = false;
@@ -132,14 +133,16 @@ public class Instrumentation {
                         .build()))
             .build();
     LogEntry entry =
-        LogEntry.of(
-            JsonPayload.of(
-                Struct.newBuilder()
-                    .putAllFields(
-                        ImmutableMap.of(
-                            DIAGNOSTIC_INFO_KEY,
-                            Value.newBuilder().setStructValue(instrumentation).build()))
-                    .build()));
+        LogEntry.newBuilder(
+                JsonPayload.of(
+                    Struct.newBuilder()
+                        .putAllFields(
+                            ImmutableMap.of(
+                                DIAGNOSTIC_INFO_KEY,
+                                Value.newBuilder().setStructValue(instrumentation).build()))
+                        .build()))
+            .setLogName(INSTRUMENTATION_LOG_NAME)
+            .build();
     return entry;
   }
 
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/Instrumentation.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/Instrumentation.java
@@ -1,0 +1,183 @@
+/*
+ * Copyright 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.logging;
+
+import com.google.api.client.util.Strings;
+import com.google.api.gax.core.GaxProperties;
+import com.google.cloud.Tuple;
+import com.google.cloud.logging.Payload.JsonPayload;
+import com.google.cloud.logging.Payload.Type;
+import com.google.common.collect.ImmutableMap;
+import com.google.protobuf.ListValue;
+import com.google.protobuf.Struct;
+import com.google.protobuf.Value;
+import java.util.ArrayList;
+import java.util.List;
+
+public class Instrumentation {
+  public static final String DIAGNOSTIC_INFO_KEY = "logging.googleapis.com/diagnostic";
+  public static final String INSTRUMENTATION_SOURCE_KEY = "instrumentation_source";
+  public static final String INSTRUMENTATION_NAME_KEY = "name";
+  public static final String INSTRUMENTATION_VERSION_KEY = "version";
+  public static final String JAVA_LIBRARY_NAME_PREFIX = "java";
+  public static final String DEFAULT_INSTRUMENTATION_VERSION = "UNKNOWN";
+  public static final int MAX_DIAGNOSTIC_VALUE_LENGTH = 14;
+  public static final int MAX_DIAGNOSTIC_ENTIES = 3;
+  private static boolean instrumentationAdded = false;
+
+  public static Tuple<Boolean, Iterable<LogEntry>> populateInstrumentationInfo(
+      Iterable<LogEntry> logEntries) {
+    boolean isWritten = setInstrumentationStatus(true);
+    List<LogEntry> entries = new ArrayList<>();
+    boolean isInfoAdded = false;
+    for (LogEntry logEntry : logEntries) {
+      // Check if LogEntry has a proper payload and also contains a diagnostic entry
+      if (logEntry.getPayload().getType() == Type.JSON
+          && logEntry
+              .<Payload.JsonPayload>getPayload()
+              .getData()
+              .containsFields(DIAGNOSTIC_INFO_KEY)) {
+        try {
+          ListValue infoList =
+              logEntry
+                  .<Payload.JsonPayload>getPayload()
+                  .getData()
+                  .getFieldsOrThrow(DIAGNOSTIC_INFO_KEY)
+                  .getStructValue()
+                  .getFieldsOrThrow(INSTRUMENTATION_SOURCE_KEY)
+                  .getListValue();
+          entries.add(createDiagnosticEntry(null, null, infoList));
+          isInfoAdded = isWritten = true;
+        } catch (Exception ex) {
+          System.err.println("ERROR: unexpected exception in populateInstrumentationInfo: " + ex);
+        }
+      } else {
+        entries.add(logEntry);
+      }
+    }
+    if (!isWritten) {
+      entries.add(createDiagnosticEntry(null, null, null));
+      isInfoAdded = true;
+    }
+    return Tuple.of(isInfoAdded, entries);
+  }
+
+  /**
+   * The helper method to generate a log entry with diagnostic instrumentation data.
+   *
+   * @param libraryName {string} The name of the logging library to be reported. Should be prefixed
+   *     with 'java'. Will be truncated if longer than 14 characters.
+   * @param libraryVersion {string} The version of the logging library to be reported. Will be
+   *     truncated if longer than 14 characters.
+   * @returns {LogEntry} The entry with diagnostic instrumentation data.
+   */
+  public static LogEntry createDiagnosticEntry(String libraryName, String libraryVersion) {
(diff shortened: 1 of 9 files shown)


google-cloud-logging/src/main/java/com/google/cloud/logging/Instrumentation.java
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  /**
+   * The helper method used to set a status of a flag which indicates if instrumentation info
+   * already written or not.
+   *
+   * @param value {boolean} The value to be set.
+   * @returns The value of the flag before it is set.
+   */
+  public static boolean setInstrumentationStatus(boolean value) {
```
minherz, line 168: nit: Use of two methods seems unnecessary complex and the naming can be confusing to a reader. It is possible to use a synchronized block of code with some object:
```java
synchronized(mutexObject) {
  // your code here
}
```
The object can be a variable of type `Boolean` (but not `boolean` value).

google-cloud-logging/src/main/java/com/google/cloud/logging/Instrumentation.java
```diff
@@ -0,0 +1,196 @@
...
+                val.getStructValue().getFieldsOrThrow(INSTRUMENTATION_NAME_KEY).getStringValue();
+            if (Strings.isNullOrEmpty(name) || !name.startsWith(JAVA_LIBRARY_NAME_PREFIX)) continue;
+            String version =
+                val.getStructValue().getFieldsOrThrow(INSTRUMENTATION_VERSION_KEY).getStringValue();
+            if (Strings.isNullOrEmpty(version)) continue;
+            libraryList.addValues(
+                Value.newBuilder().setStructValue(createInfoStruct(name, version)).build());
+            if (libraryList.getValuesCount() == MAX_DIAGNOSTIC_ENTIES) break;
+          } catch (Exception ex) {
+            System.err.println("ERROR: unexpected exception in generateLibrariesList: " + ex);
```
minherz, line 142: nit: while we print errors to `stderr` we do it in circumstances when the error is critical. i would consider just skipping the problematic entry and avoid yielding unexpected output. if you decide to keep printing the error, please make the error message more verbose, so a developer will be able to discover the source of the error.

google-cloud-logging/src/main/java/com/google/cloud/logging/Instrumentation.java
```diff
@@ -0,0 +1,196 @@
...
+   * @param libraryVersion {string} The version of the logging library to be reported. Will be
+   *     truncated if longer than 14 characters.
+   * @returns {LogEntry} The entry with diagnostic instrumentation data.
+   */
+  public static LogEntry createDiagnosticEntry(String libraryName, String libraryVersion) {
+    return createDiagnosticEntry(libraryName, libraryVersion, null);
+  }
+
+  private static LogEntry createDiagnosticEntry(
+      String libraryName, String libraryVersion, ListValue existingLibraryList) {
```
minherz, line 94: nit: the code that creates protobuf entry is hard to read. would it be possible to create the `Dictionary<?,?>` based payload? i think the code will be more readable and further developers can easily guess the instrumentation payload structure.

google-cloud-logging/src/main/java/com/google/cloud/logging/Instrumentation.java
```diff
@@ -0,0 +1,196 @@
...
+  public static final String INSTRUMENTATION_SOURCE_KEY = "instrumentation_source";
+  public static final String INSTRUMENTATION_NAME_KEY = "name";
+  public static final String INSTRUMENTATION_VERSION_KEY = "version";
+  public static final String JAVA_LIBRARY_NAME_PREFIX = "java";
+  public static final String DEFAULT_INSTRUMENTATION_VERSION = "UNKNOWN";
+  public static final int MAX_DIAGNOSTIC_VALUE_LENGTH = 14;
+  public static final int MAX_DIAGNOSTIC_ENTIES = 3;
+  private static boolean instrumentationAdded = false;
+
+  public static Tuple<Boolean, Iterable<LogEntry>> populateInstrumentationInfo(
```
minherz, line 42: nit: consider declaring a class instead of  the `Tuple<?,?>`. The `x()` and `y()` calls in the `LoggingHandler.java` cannot be guessed without reading through all code. Method documentation can be added to elaborate about the returned values.

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -321,7 +324,15 @@ public void publish(LogRecord record) {
...
-          logging.write(logEntries, defaultWriteOptions);
+          // Add partialSuccess option always for request containing instrumentation data
+          if (pair.x()) {
+            List<WriteOption> writeOptions = new ArrayList<WriteOption>();
+            writeOptions.addAll(Arrays.asList(defaultWriteOptions));
+            writeOptions.add(WriteOption.partialSuccess(true));
+            logging.write(logEntries, Iterables.toArray(writeOptions, WriteOption.class));
+          } else {
+            logging.write(logEntries, defaultWriteOptions);
+          }
```
minherz, line 335: nit: the flag can always be added to avoid if/else: `WriteOption.partialSuccess(pair.x())`

minherz (commented):
I did not find instrumentation for the calls to `Logging.Write()` and `Logging.WriteAsync()`. Are they expected to be in a separate PR? I recommend to instrument all paths before merging.

google-cloud-logging/src/main/java/com/google/cloud/logging/Instrumentation.java
losalex: Thanks for input! I like the tuple definition since it does not hides the values it carries and also this code is not on the main path of exposed APIs. Thats said I went ahead and documented the API signature

google-cloud-logging/src/main/java/com/google/cloud/logging/Instrumentation.java
losalex: Sure - I can remove it

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
losalex: I reworked this part of code - let me know WDYT. I was trying to avoid adding partialSuccess always to make sure we do not break existing tests

google-cloud-logging/src/main/java/com/google/cloud/logging/Instrumentation.java
losalex: Sure - added a separate `Object` to be used for synchronization

google-cloud-logging/src/main/java/com/google/cloud/logging/Instrumentation.java
losalex: Sure - added a separate `Object` to be used for synchronization

google-cloud-logging/src/main/java/com/google/cloud/logging/Instrumentation.java
losalex: > I did not find instrumentation for the calls to `Logging.Write()` and `Logging.WriteAsync()`. Are they expected to be in a separate PR? I recommend to instrument all paths before merging.

Right - good catch! Eventually `Logging.writeAsync()` is private, but `Logging.write()` is public and I missed it

losalex: > nit: the code that creates protobuf entry is hard to read. would it be possible to create the `Dictionary<?,?>` based payload? i think the code will be more readable and further developers can easily guess the instrumentation payload structure.

To be honest, I prefer the protobuf structure since it is safer from types handling perspective. If you look into code I added, it uses bunch of useful type checking methods like` hasStructValue()` or strongly types setters/getters like `setStructValue()` /  `getValuesList()`. Also I liked methods as `getFieldsOrThrow()` which pretty convenient. Of course I could achieve type casting of `Map<String, Object>` values, but it would make code to be complex in other parts like reading...

google-cloud-logging/src/main/java/com/google/cloud/logging/Instrumentation.java
minherz: I think by _hiding_ you reference the visual hints that IDE (such as VSCode) expose when working with tuple objects. While it is true that you get hints about the type, it cannot replace the additional info that is conveyed via the good variable/function/class name. Leave it at your discretion though.

google-cloud-logging/src/main/java/com/google/cloud/logging/Instrumentation.java
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  /**
+   * The helper method used to set a status of a flag which indicates if instrumentation info
+   * already written or not.
+   *
+   * @param value {boolean} The value to be set.
+   * @returns The value of the flag before it was set.
+   */
+  public static boolean setInstrumentationStatus(boolean value) {
```
minherz, line 195: nit: instrumentation statuses true and false may not be clear for a reader who is not familiar with the subject. consider replacing the method name or use `enum` as an argument or constants for calling the method.

minherz (approved):
there is a case of NPE that should be handled. Left few more comments for considerations. Beside that it is good to go.

google-cloud-logging/src/main/java/com/google/cloud/logging/Instrumentation.java
losalex: The variable names could be indeed informative, but creating a class just for grouping returned values could be a bit too much - thats what is `Tuple` designed for. Given a fact that this functionality is not supposed to be used publicly by developers, I think `Tuple` is exactly what we need - I wish we could have an anonymous types or anything similar in Java

google-cloud-logging/src/main/java/com/google/cloud/logging/Instrumentation.java
```diff
@@ -0,0 +1,221 @@
...
+                        ImmutableMap.of(
+                            DIAGNOSTIC_INFO_KEY,
+                            Value.newBuilder().setStructValue(instrumentation).build()))
+                    .build()));
+    return entry;
+  }
+
+  private static ListValue generateLibrariesList(
+      String libraryName, String libraryVersion, ListValue existingLibraryList) {
+    if (Strings.isNullOrEmpty(libraryName) || !libraryName.startsWith(JAVA_LIBRARY_NAME_PREFIX))
```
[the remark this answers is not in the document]
losalex: This is going to be used by other libraries to provide their own name/version - they will call to `createDiagnosticEntry() `with those parameters.

google-cloud-logging/src/main/java/com/google/cloud/logging/Instrumentation.java
```diff
@@ -0,0 +1,221 @@
...
+      libraryName = JAVA_LIBRARY_NAME_PREFIX;
+    if (Strings.isNullOrEmpty(libraryVersion)) {
+      libraryVersion = getLibraryVersion(Instrumentation.class.getClass());
+    }
+    Struct libraryInfo = createInfoStruct(libraryName, libraryVersion);
+    ListValue.Builder libraryList = ListValue.newBuilder();
+    // Append first the library info for this library
+    libraryList.addValues(Value.newBuilder().setStructValue(libraryInfo).build());
+    if (existingLibraryList != null) {
+      for (Value val : existingLibraryList.getValuesList()) {
```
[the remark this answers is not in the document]
losalex: As I mentioned earlier, I prefer to use APIs like `hasStructValue()`, `getStructValue()`, `getFieldsOrThrow()` and not to start making casting to specific objects from dictionary which clearly will also add code complexity and potentially could be dangerous

google-cloud-logging/src/main/java/com/google/cloud/logging/Instrumentation.java
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  /**
+   * Adds a partialSuccess flag option to array of WriteOption
+   *
+   * @param options {WriteOption[]} The options array to be extended
+   * @return The new array of oprions containing WriteOption.OptionType.PARTIAL_SUCCESS flag set to
+   *     true
+   */
+  public static WriteOption[] addPartialSuccessOption(WriteOption[] options) {
```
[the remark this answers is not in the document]
losalex: `options` are turned into map in `write()` method, but for the sake of proper input handling I will add the check for null - thanks!

google-cloud-logging/src/main/java/com/google/cloud/logging/Instrumentation.java
losalex: I reworked this part - I made `setInstrumentationStatus()` only to set a flag and added package internal method `resetInstrumentationStatus()` to reset the flag

PR #983, opened 2022-06-27
chore: mark native image checks as required

mpeddada1:


losalex (commented):
I think we cannot enable those checks to be mandatory by default unless we add missing integration tests

mpeddada1:
Currently running into: 
```
com.google.cloud.logging.LoggingException: io.grpc.StatusRuntimeException: RESOURCE_EXHAUSTED: Quota exceeded for quota metric 'Read requests' and limit 'Read requests per minute' of service 'logging.googleapis.com' for consumer 'project_number:1016721519174'.
```

losalex:
> Currently running into:
> 
> ```
> com.google.cloud.logging.LoggingException: io.grpc.StatusRuntimeException: RESOURCE_EXHAUSTED: Quota exceeded for quota metric 'Read requests' and limit 'Read requests per minute' of service 'logging.googleapis.com' for consumer 'project_number:1016721519174'.
> ```

We need to wait and retry - unfortunately we must make sure that step "Kokoro - Test: Java 17 GraalVM Native Image" is succeeding

mpeddada1:
Got it, thanks @losalex. Looks like rerunning the jobs did the trick.

losalex (approved):

PR #1006, opened 2022-07-23
fix: Fix flakiness in JSON test

### LINKED ISSUE #1005: LogEntryTest.java#testStructureLogPresentations is flaky
LogEntryTest.java#testStructureLogPresentations is rendering a LogEntry into JSON and comparing it to a literal string.

This is flaky because JSON does not guarantee dictionary element ordering, and the order indeed does change (e.g., in my local tests).

Will send a PR to fix

bpcreech:
Fixes #1005


losalex (approved):
LGTM

PR #1010, opened 2022-07-23
fix: Apply fixes from Google error-prone

### LINKED ISSUE #1009: Various Google error-prone findings
Running the most recent Google error-prone over this codebase turns up a few categories of issues, e.g.:

1. (New) Lack of notation of `CanIgnoreReturnValue` on methods which return this (typically Builder methods). This partially inhibits effective use of `CheckReturnValue` by users of this library.
2. A few methods that aren't marked `@Nullable` but clearly return null. This partially inhitibs effective use of tools like `NullAway`.
3. `getClass` called on `.class`, which is redundant
4. Attempts to check for proto lists using `== null`. Proto fields are never null, these should check `.isEmpty()`.
5. A handful of tests that aren't running because of no `@Test` annotation (one initially commented out, but that one should use `@Ignore`... and another test which is not enabled and also doesn't seem to test anything [`testFromCurrentContextWithNegativeLevel`])
6. Backwards `assertEqual(x, y)` statements (which yields backwards error messages on failure)
7. One unnecessary `@SupressWarnings("unchecked")`
8. A few more tiny things

Will send a PR

bpcreech:
Fixes #1009

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,20 +51,20 @@ If you are using Maven without BOM, add this to your dependencies:
 If you are using Gradle 5.x or later, add this to your dependencies:
 
 ```Groovy
-implementation platform('com.google.cloud:libraries-bom:26.0.0')
+implementation platform('com.google.cloud:libraries-bom:26.1.0')
 
 implementation 'com.google.cloud:google-cloud-logging'
 ```
 If you are using Gradle without BOM, add this to your dependencies:
 
 ```Groovy
-implementation 'com.google.cloud:google-cloud-logging:3.10.0'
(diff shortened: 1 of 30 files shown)


losalex (approved):
LGTM

losalex (commented):
Seems the cli check fails with following error:

Warning:  Used undeclared dependencies found:
Warning:     com.google.errorprone:error_prone_annotations:jar:2.14.0:compile

Perhaps you can take a look?

bpcreech:
Argh, my bad, I didn't run the dependencies check on this and thus didn't add the error-prone annotations to pom.xml, and the test failed. Fixed!

PR #1012, opened 2022-07-24
fix: Replace Date with LocalDateTime

### LINKED ISSUE #1011: Let's use java.time.LocalDateTime instead of java.util.Date
`java.time.LocalDateTime` is generally the preferred DateTime handling library since Java 8, due to various footguns in the old `java.util.Date` interface.

I'll send a PR for a few instances this.

bpcreech:
Fixes #1011

### DIFF
--- a/google-cloud-logging/src/test/java/com/google/cloud/logging/TimestampDefaultFilterTest.java
+++ b/google-cloud-logging/src/test/java/com/google/cloud/logging/TimestampDefaultFilterTest.java
@@ -16,15 +16,13 @@
 
 package com.google.cloud.logging;
 
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
+import static com.google.common.truth.Truth.assertThat;
+import static java.time.ZoneOffset.UTC;
+import static java.util.Locale.US;
 
-import java.text.DateFormat;
-import java.text.SimpleDateFormat;
-import java.util.Calendar;
-import java.util.Date;
-import java.util.TimeZone;
-import javax.management.timer.Timer;
+import java.time.Duration;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
 import org.junit.Test;
 
 public class TimestampDefaultFilterTest {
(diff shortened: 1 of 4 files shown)


google-cloud-logging/src/test/java/com/google/cloud/logging/TimestampDefaultFilterTest.java
```diff
@@ -33,23 +31,20 @@ public class TimestampDefaultFilterTest {
...
-    try {
-      DateFormat rfcDateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'");
-      rfcDateFormat.setTimeZone(TimeZone.getTimeZone("UTC"));
-      Date actual = rfcDateFormat.parse(defaultFilter.substring(12, defaultFilter.length() - 1));
-      assertTrue(Math.abs(expected.getTime() - actual.getTime()) < Timer.ONE_MINUTE);
-    } catch (java.text.ParseException ex) {
-      fail(); // Just fail if exception is thrown
-    }
+    DateTimeFormatter rfcDateFormat =
+        DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'", US);
```
losalex, line 40: Should it be UTC?

losalex (approved):
LGTM - left a comment, please let me know if it is not a behavior change

google-cloud-logging/src/test/java/com/google/cloud/logging/TimestampDefaultFilterTest.java
bpcreech: [This is Locale, not timezone](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#ofPattern-java.lang.String-). Locale would be used if the pattern had something Locale-specific, like a month name. AFAICT there aren't any Locale-dependent components in this parse string so it's kind of a no-op, but nonetheless our linter encourages us to always specify local because otherwise it's read from the environment, which makes the system unnecessarily environment-dependent. (This timestamp is being generated as a Cloud Logging filter so it should always use the same Locale, US, to the extent that it matters.)

Meanwhile timezone is still actually being read from the input here; it's the `Z` part of the parse format string.

bpcreech:
Ah, my bad, had a trivial bug in this, hadn't run the integration tests. (Actually just a bug in the test FWIW.) Was trying to get a datetime format out of a `LocalDate`; needs a `LocalDateTime`.

It's fixed now.

PR #1014, opened 2022-07-24
fix: Supress unchecked conversion warnings

### LINKED ISSUE #1013: Payload.fromPb and Destination.fromPb method overrides emit unchecked conversion warnings
Compiling with Java 11 (not sure about other Javas), the overrides of `Payload.fromPb` and `Destination.fromPb` emit unchecked conversion warnings, because they're all redefining the return type of the overridden method. E.g., [this](https://stackoverflow.com/questions/23645753/java-unchecked-overriding-return-type)

This setup seems like a design tenet of this library. Um, it might be nice to redesign to not change return type in a child class, but in the meanwhile, I guess we should suppress unchecked conversion warnings since this is intentional.

I'll send a PR.

bpcreech:
Fixes #1013

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,20 +51,20 @@ If you are using Maven without BOM, add this to your dependencies:
 If you are using Gradle 5.x or later, add this to your dependencies:
 
 ```Groovy
-implementation platform('com.google.cloud:libraries-bom:26.0.0')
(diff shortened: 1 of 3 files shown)


losalex (approved):
LGTM

PR #1020, opened 2022-07-30
fix: Correct the JSON time field name

### LINKED ISSUE #1019: JSON formatter uses wrong time field
`LogEntry.StructuredLogFormatter.toStructuredJsonString` is writing to the wrong time field (or writing the wrong data to it).

[The spec](https://cloud.google.com/logging/docs/agent/logging/configuration#timestamp-processing) states we should use:

 * `timestamp` with a structure of epoch `seconds` and `nanos`, *or*
 * `time` with RFC3339 time.

Currently `LogEntry.StructuredLogFormatter.toStructuredJsonString` [is writing](https://github.com/googleapis/java-logging/blob/4c597dce9f6da92ffb81e753aed4a1fe584ba365/google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java#L703) RFC3339 time to the `timestamp` field.

This breaks our parser for our test harness. I'm not sure if it works in all GCP properties which use this format. (Generally the behavior of logging parsers is to fall back to current time, so I would be unsurprised if this is silently failing down to slightly less accurate timestamps in places.)

I'll send a PR.

bpcreech:
Fixes #1019


google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
```diff
@@ -700,7 +700,7 @@ public String toStructuredJsonString() {
 
     formatter
         .appendField("severity", severity)
-        .appendField("timestamp", timestamp)
+        .appendField("time", timestamp)
```
losalex, line 703: Such change could break existing clients which use `timestamp` in logs. I believe to be on a safe side we should add a configuration parameter controlling this behavior with default to be working with `timestamp`.

bpcreech: Unfortunately I think anyone using this library is already broken. I don't think this code is likely to be working for anyone as-is. Unless they implemented their own parser for some reason.

[Here is the spec for these fields](https://cloud.google.com/logging/docs/agent/logging/configuration#timestamp-processing). Note that this code has been putting RFC3339 time (what happens when you implicitly call toString on an Instant) into the timestamp field. That is wrong.

[https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/blob/3191146fb98c55dd5e8a2f91ef491b52d09503aa/lib/fluent/plugin/out_google_cloud.rb#L1322](Here is the code in the Google fluentd parser) that parses this format. This is the original and primary intended parser for this format. Note that it doesn't know what to do if the `timestamp` field has a string in it. Instead it's defaulting to current time, meaning that everyone using this library with the Google Fluentd Parser is getting less accurate timestamps, representing when the parser sees the log as opposed to when the call site emitted it.

The other primary use case for this format (that I know of anyway) is GCP Serverless. Specifically this format works on App Engine Standard gen 2 and Cloud Run Managed. Internally, that code lives at go/time-is-broken. You'll note that if the field called `timestamp` is a string, it's ignored.

If we leave this broken default, it means this library will be broken by default for anyone trying to use it with the Google Fluentd agent, or GCP Serverless. :(

losalex: Makes sense

losalex (approved):
LGTM

Issue #747, opened 2021-11-14
feat: Enable access to LogRecord in LoggingEnhancer (e.g., for access to flogger LogSite)


bpcreech:
**Is your feature request related to a problem? Please describe.**

The `LoggingEnhancer` interface enables an implementer to add extra content to a `LogEntry`. However, the `LoggingEnhancer` is not itself given any context about the underlying `LogRecord`, so any added content must essentially be static and unrelated to the specific log record.

Some logging systems, like flogger, subclass `java.util.logging.LogRecord` (e.g., flogger makes a `com.google.common.flogger.backend.system.AbstractLogRecord`) and add extra information (e.g., in the case of flogger, the `AbstractLogRecord` has `getLogData().getLogSite()` which offers a much more efficient and less hassle-free mechanism of retrieving log site information than `com.google.cloud.logging.SourceLocation.fromCurrentContext`).

**Describe the solution you'd like**

Can we add a `LogRecord` parameter to `LoggingEnhancer.enhanceLogEntry`? This would be a breaking change; if it's too late for such a change, maybe we need a `LoggingEnhancerWithRecord` or `LoggingEnhancer2`.

Then as a follow-on, some library can supply the glue for flogger and this library: a stock `LoggingEnhancer` to supply the `LogSite`.

**Describe alternatives you've considered**

We currently have a logging handler which downcasts the `LogRecord`, but it has its own `java.util.logging.LogHandler` implementation and thus doesn't use `LoggingHandler` or the  `java-logging` library. I'm filing this feature request in hopes of replacing that internal implementation.

**Additional context**

Contact me for the internal example of downcasting `LogRecord`. :)

minherz:
Hi, the information from the LogRecord is translated to LogEntry:
https://github.com/googleapis/java-logging/blob/c003417aade41e025082f47f2185e66707bffcff/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java#L268-L295

If there is any information that is not captured, a user can develop an enhancer and register it to enhance resulted LogEntry instance. It will help us to understand better your request if you can provide an example of the data that you are trying to log with JUL and what data you eventually expect to see in Log Explorer.

bpcreech:
Right the problem is that the enhancer doesn't get any access to the LogRecord.

[...]

Issue #32, opened 2019-06-05
Make `LoggingEventEnhancer` more generic so can be used in more implementations 


dansiviter:
**Is your feature request related to a problem? Please describe.**
Currently only `com.google.cloud.logging.logback.LoggingAppender` can enhance using information contained within the logging event. This isn't possible via the JDK logger although it should be.

 **Describe the solution you'd like**
Replace `com.google.cloud.logging.logback.LoggingEventEnhancer` with `com.google.cloud.logging.LoggingEventEnhancer` that has a class structure similar to:

```
@FunctionalInterface
public interface LoggingEventEnhancer<E> {
  void enhanceLogEntry(LogEntry.Builder builder, E event);
}
```

For backwards compatibility `com.google.cloud.logging.logback.LoggingEventEnhancer` could then change to:
```
public interface LoggingEventEnhancer extends com.google.cloud.logging.LoggingEventEnhancer<ILoggingEvent> {
   ...
}
```

 **Describe alternatives you've considered**
None

 **Additional context**
The JDK logger `com.google.cloud.logging.LoggingHandler` should then be enhanced to perform the enhancement using the new class. The old `com.google.cloud.logging.LoggingEnhancer` could be deprecated as its functionality could be subsumed by the new one.

minherz:
@dansiviter do you still get these errors?

minherz:
Thank you for your proposal. Indeed it enhances the abilities to customize the published logs by introducing a common logic across the application. We consider this change to be a breaking change. It cannot be implemented without either creating a flow that duplicates the existing method or by changing the existing `LoggingEnhancer` interface. It means this change can be introduced only within the next major release.
Please, see if the next release (probably 3.6) that introducing automatic metadata population helps you with some of the tasks for automatic log entry enhancement. See PR #808 for more details.

losalex:
I am closing this issue since seems it is pretty old and apparently seems there is no much demand for this feature - please let me know if anyone thinks otherwise

PR #1021, opened 2022-07-30
fix: Make LoggingHandler.logEntryFor extensible

bpcreech:
This is a stopgap solution suggested by minherz@ in #747. The long-term fix is #32 but that seems like it could take some time (it's been open 3 years and is pending a major version bump).

This enables us to enhance the `c.g.cloud.logging.LogEntry` in ways that require access to the original `java.util.logging.LogRecord`, e.g.:

 * Setting `SourceLocation` using extra, more accurate context the Google Flogger library captures (on a subclass of the JUL `LogRecord`),
 * Recording `logRecord.getThreadID`,
 * Recording exception name explicitly from `record.getThrown()`,
 * etc

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,20 +51,20 @@ If you are using Maven without BOM, add this to your dependencies:
 If you are using Gradle 5.x or later, add this to your dependencies:
 
 ```Groovy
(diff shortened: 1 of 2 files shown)


losalex (approved):
LGTM

PR #1027, opened 2022-08-06
feat: Add support for JSON logging to stderr

### LINKED ISSUE #1026: Support JSON logging to stderr
The `LoggingHandler` currently has a mode to log JSON to stdout, using [the structured logging](https://cloud.google.com/logging/docs/structured-logging) format.

In constrast, many organizations standardize on stderr for logging, with stdout being reserved ["for regular output"](https://stackoverflow.com/questions/4919093/should-i-log-messages-to-stderr-or-stdout). I work for such an organization. :) Of course, one could argue that the java-logging library is primarily for servers which *have no regular output*, but nonetheless it seems useful to add a stderr option onto this library.

I'll send a PR for this.

bpcreech:
Fixes #1026

### DIFF
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
@@ -115,10 +115,15 @@ import org.jspecify.nullness.Nullable;
  *       Logging (defaults to {@code true}).
  *   <li>{@code com.google.cloud.logging.LoggingHandler.redirectToStdout} is a boolean flag that
  *       opts-in redirecting the output of the handler to STDOUT instead of ingesting logs to Cloud
- *       Logging using Logging API (defaults to {@code true}). Redirecting logs can be used in
+ *       Logging using Logging API (defaults to {@code false}). Redirecting logs can be used in
  *       Google Cloud environments with installed logging agent to delegate log ingestions to the
  *       agent. Redirected logs are formatted as one line Json string following the structured
- *       logging guidelines.
+ *       logging guidelines. This flag is deprecated; use {@code
+ *       com.google.cloud.logging.LoggingHandler.logTarget} instead.
+ *   <li>{@code com.google.cloud.logging.LoggingHandler.logTarget} is an enumeration controlling log
+ *       routing (defaults to {@code CLOUD_LOGGING}). If set to STDOUT or STDERR, logs will be
+ *       printed to the corresponding stream in the Json format that can be parsed by the logging
+ *       agent. If set to CLOUD_LOGGING, logs will be sent directly to the Google Cloud Logging API.
  * </ul>
  *
  * <p>To add a {@code LoggingHandler} to an existing {@link Logger} and be sure to avoid infinite
@@ -136,6 +141,16 @@ public class LoggingHandler extends Handler {
   private static final String LEVEL_NAME_KEY = "levelName";
   private static final String LEVEL_VALUE_KEY = "levelValue";
 
+  /** Where to send logs. */
+  public enum LogTarget {
+    /** Sends logs to the Cloud Logging API. */
+    CLOUD_LOGGING,
+    /** Sends JSON-formatted logs to stdout, for use with the Google Cloud logging agent. */
+    STDOUT,
+    /** Sends JSON-formatted logs to stderr, for use with the Google Cloud logging agent. */
+    STDERR
+  }
+
   private final List<LoggingEnhancer> enhancers;
   private final LoggingOptions loggingOptions;
 
@@ -152,7 +167,7 @@ public class LoggingHandler extends Handler {
   private volatile Level flushLevel;
 
   private volatile Boolean autoPopulateMetadata;
-  private volatile Boolean redirectToStdout;
+  private volatile LogTarget logTarget;
 
   private final WriteOption[] defaultWriteOptions;
 
@@ -243,7 +258,13 @@ public class LoggingHandler extends Handler {
       Boolean f1 = loggingOptions.getAutoPopulateMetadata();
       Boolean f2 = config.getAutoPopulateMetadata();
       autoPopulateMetadata = isTrueOrNull(f1) && isTrueOrNull(f2);
-      redirectToStdout = firstNonNull(config.getRedirectToStdout(), Boolean.FALSE);
+      logTarget =
+          config
+              .getLogTarget()
+              .orElse(
+                  firstNonNull(config.getRedirectToStdout(), Boolean.FALSE)
+                      ? LogTarget.STDOUT
+                      : LogTarget.CLOUD_LOGGING);
       String logName = log != null ? log : config.getLogName();
       MonitoredResource resource =
           firstNonNull(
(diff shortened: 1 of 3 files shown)


losalex (commented):
@bpcreech , thanks for opening this PR! I have a conceptual feedback on this change - basically when you configure `stdout`/`stderr` to be used, it will redirect **all** logging traffic to `stdout`/`stderr` respectively. While developer always can create 2 loggers to be able to use both, `stdout` and `stderr` streams, it would be best to make this feature dependent on [Severity](https://github.com/googleapis/java-logging/blob/9d70b96914ae6217cbe58ab1af93e1d5dfe67ce1/google-cloud-logging/src/main/java/com/google/cloud/logging/Severity.java#L27). This way we can define more flexible logic to send log record to dedicated output streams based if the record has error-related severity or not (e.g. DEFAULT/DEBUG/INFO/NOTICE severities will go to `stdout` and the rest to `stderr`).
Thats said, I believe that `Severity` based output could be introduced as separate feature in addition to ability to redirect everything to `stdout` or `stderr` - please let me know WDYT about this and we can take it from there.

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -428,13 +455,33 @@ public Boolean getAutoPopulateMetadata() {
...
-    return redirectToStdout;
+    return this.logTarget == LogTarget.STDOUT;
+  }
+
+  /**
+   * Configure the destination for ingested logs. If set to STDOUT or STDERR, logs will be printed
+   * to the corresponding stream in the Json format that can be parsed by the logging agent. If set
+   * to CLOUD_LOGGING, logs will be sent directly to the Google Cloud Logging API.
+   */
+  public void setLogTarget(LogTarget value) {
```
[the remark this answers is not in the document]
bpcreech: I think generally sending logs to *both* stdout and stderr would be unusual. I think the most common consensus e.g., [on StackOverflow](https://stackoverflow.com/questions/4919093/should-i-log-messages-to-stderr-or-stdout) is that logs should *only* go to `stderr`; `stdout` is for the program's primary output, if it has it. I've also [seen](https://github.com/bloomberg/bde/blob/master/groups/bal/ball/ball_fileobserver.h) stderr or stdout + file with severity-based thresholds for each. But doing both stderr and stdout seems unusual.

In my opinion to avoid unnecessary complexity in the interface, it's best to tell anyone who wants some permutation of (stdout, stderr, file, Cloud Logging) based on severity to *instantiate two handlers*. You can already express this in JUL; it has built-in systems to manage the log multiplexing. I think building multiplexing into the handler is a mistake that, e.g., leads one to later simplify the [two-output handler](https://github.com/bloomberg/bde/blob/master/groups/bal/ball/ball_fileobserver.h) down to [a simple one-output one](https://github.com/bloomberg/bde/blob/master/groups/bal/ball/ball_fileobserver2.h).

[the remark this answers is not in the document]
bpcreech: ... Regarding adding `setRedirectToStderr`: This is somewhat hinges on decision of whether we want to build multiplexing in the `LoggingHandler`. IMO we shouldn't, a LoggingHandler instance should have one output at a time.

If we do that, the output control is a single variable: an enum, not a boolean. Because of that in this PR I **deprecated** `setRedirectToStdout` since it takes a boolean, not the new `LogTarget` enum. Adding `setRedirectToStderr` would add another already-deprecated method.

Of course we could continue to *not do multiplexing in `LoggingHandler`*, while also leaving ``setRedirectToStdout` **un**deprecated, and add `setRedirectToStdout`... they'd just be semantically misleading since they'd be acting like setters for two booleans when they're actually setters on the same enum.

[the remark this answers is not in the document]
losalex: > ... Regarding adding `setRedirectToStderr`: This is somewhat hinges on decision of whether we want to build multiplexing in the `LoggingHandler`. IMO we shouldn't, a LoggingHandler instance should have one output at a time.
> 
> If we do that, the output control is a single variable: an enum, not a boolean. Because of that in this PR I **deprecated** `setRedirectToStdout` since it takes a boolean, not the new `LogTarget` enum. Adding `setRedirectToStderr` would add another already-deprecated method.
> 
> Of course we could continue to _not do multiplexing in `LoggingHandler`_, while also leaving `setRedirectToStdout` **un**deprecated, and add `setRedirectToStdout`... they'd just be semantically misleading since they'd be acting like setters for two booleans when they're actually setters on the same enum.

Yep, this is indeed a bit unusual... I must admit that `setLogTarget()` is confusing as well - library user might not understand immediately that `setLogTarget()` and `setRedirectToStdout()` are mutually exclusive...

losalex (approved):

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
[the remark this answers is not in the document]
bpcreech: Okay, added an explicit comment to both `setLogTarget` and `setRedirectToStdout` that they're mutually exclusive.

I'd already marked `setRedirectToStdout` as deprecated.

nickkieffer:
Hey @losalex and @bpcreech, I think this change made it so the default python logging package no longer logs to gcloud without using a structured log to STOUT
I think it was logging to STERR by default previously (yes malformed but it still worked)
This was before the release as an example
image
Now we just get no logs, working to add these to STOUT and json structured but figured others will run into this

bpcreech:
Hmm, this change was to the Java logging library, which is not used by any Python SDK I'm aware of.

nickkieffer:
I was thinking it may consume logs that are formatted via the python SDK, nothing directly related if that makes sense

bpcreech:
Can you describe your software stack? What environment are you running on, and what SDKs are you using?

nickkieffer:
We run python scripts using kubernetes in gcloud,
in python just the basic
```
import logging
log = logging.getLogger(__name__)
log.debug('debuggin')
```

This is pretty confusing to me though, google stackdriver doesn't have our logs but 
when using the kubernetes python SDK we actually DO get logs back so we know we are logging something still 😰 

```
from kubernetes.client import BatchV1Api, LogsApi, CoreV1Api, EventsV1beta1Api
api = api_mgr.get(CoreV1Api)
pod = api.read_namespaced_pod(pod_name, self.namespace)
logs = api.read_namespaced_pod_log(
     pod_name, self.namespace, follow=True, _preload_content=False
 )
for line in logs.stream():
            yield Event(timestamp=0, type=TYPE_POD_LOG, name=pod_name, message=line.decode('utf-8'))
```

nickkieffer:
On the left is the google console, the right is our UI displaying the logs from that kubernetes SDK
image

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -310,18 +331,24 @@ public void publish(LogRecord record) {
...
+                ? ImmutableList.of(logEntry)
+                : Instrumentation.populateInstrumentationInfo(ImmutableList.of(logEntry)).y();
         if (autoPopulateMetadata) {
           logEntries =
               logging.populateMetadata(
                   logEntries, getMonitoredResource(), "com.google.cloud.logging", "java");
         }
-        if (redirectToStdout) {
-          logEntries.forEach(log -> System.out.println(log.toStructuredJsonString()));
-        } else {
```
nickkieffer, line 323: it stands out to me that this is an if else and is replaced with a switch without a default case, but 🤷

losalex: @nickkieffer , are you using `redirectToStdout` feature?

nickkieffer: no, not directly anyways

losalex: @nickkieffer , I must admit that I have some confusion while following the issue you raised, but let me try answering best based on my understanding:

1. Correct me if I wrong, you use a Python logger - based on sample code you shared. This change to print to `STDERR` was performed only in Java library and I believe we do not have such feature in Python. If problem indeed related to Python SDK, can you please raise the issue in [Python](https://github.com/googleapis/python-logging/issues) repo? Feel free to provide all details you mentioned here and also the details about environment you run in - adding @daniel-sanche for visibility.
2. This change supposed to use `CLOUD_LOGGING` by default - the `STDOUT` and `STDERR` logging needs to be explicitly enabled in code or configuration so the change should be fully backward compatible to clients which does not use the feature. If you believe this is not a case, please let us know - me and @bpcreech can take a look.

Please let me know if I understood you correctly and if I can assist you any further.

nickkieffer: You can disregard me on this, I think my ops team just so happened to turn off of logging agents around the same time as your release happened. Sorry about the distraction!

PR #1044, opened 2022-08-24
fix: Apply Google Java Code Clarity suggestions

### LINKED ISSUE #1043: Let's apply our internal automated linter fixes
In https://github.com/googleapis/java-logging/pull/1010 we added fixes from Google Error Prone. We have a lot more internal linter fixes!

Here's a run-down of things in this repo we can automatically fix. I'll send a PR.

1. Remove a few unused imports
2. Remove or narrow declared and caught checked exceptions across the board to match what's actually possible.
3. Remove unnecessary use of Guava `Function`s and lambdas where direct method references will do.
4. Remove unnecessary use of on-the-fly class declarations where very short lambdas will do
5. Make all member variables final that we safely can.
6. Stylistically, reference inner classes (especially enums and Builders) wrt their outer class.
7. Fix javadocs which include unescaped `<` and `>`.
8. A few Javadoc fixes to make code links explicit
9. Fix botched attempt to write `@code` in `Logging.java`
10. Simplify some switch/cases (joining redundant cases and removing unnecessary default cases)
11. Switch to Immutable collections wherever it's clearly possible.
12. Add `@Override` in places where relevant
13. Add `buildOrThrow` where possible on protos
14. Replace `assertTrue` with more specific things (`assertEquals`, `assertSame`, `assertThat(X).isLessThan(Y)`, etc)
15. Remove unnecessary declaration of variables which are only returned
16. Remove unnecessary generic type specification on right-hand side of assignments.
17. Switch from LinkedList to ArrayDeque since the latter is essentially always more efficient and we're only using it as a deque.
18. Switching string splitting to `com.google.common.base.Splitter` which is more efficient.
19. Switch from `String.toLowerCase` to `Ascii.toLowerCase` for consistency in the face of different system locales
20. Switch from `String.equals("")` to `String.empty()` in a few places
21. Use `Collections.addAll(X, Y)` instead of `X.addAll(Arrays.asList(Y))` converting an array to a list.
22. Remove unnecessary class qualification in a few places
23. Make the Instrumentation class final and uninstantiatable since it only has static members.
24. Stop checking for null on non-nullable protobuf members.
25. Remove one mysterious unnecessary `String.valueOf` to make an identical string from a string.
26. Replace one `==` on a `String` with `.equals`
27. Misc other small things

bpcreech:
Fixes #1043

### DIFF
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
@@ -43,8 +43,10 @@ import com.google.cloud.PageImpl;
 import com.google.cloud.Tuple;
 import com.google.cloud.logging.spi.v2.LoggingRpc;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Ascii;
 import com.google.common.base.Function;
 import com.google.common.base.Throwables;
+import com.google.common.base.VerifyException;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
@@ -87,7 +89,9 @@ import com.google.logging.v2.WriteLogEntriesRequest;
 import com.google.logging.v2.WriteLogEntriesResponse;
 import com.google.protobuf.Empty;
 import com.google.protobuf.util.Durations;
+import java.text.ParseException;
 import java.util.*;
+import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
@@ -104,13 +108,10 @@ class LoggingImpl extends BaseService<LoggingOptions> implements Logging {
   private volatile Severity flushSeverity = null;
   private boolean closed;
 
-  private static final Function<Empty, Boolean> EMPTY_TO_BOOLEAN_FUNCTION =
-      new Function<Empty, Boolean>() {
-        @Override
-        public Boolean apply(Empty input) {
-          return input != null;
-        }
-      };
+  private static Boolean emptyToBooleanFunction(Empty input) {
+    return input != null;
+  }
+
   private static final Function<WriteLogEntriesResponse, Void> WRITE_RESPONSE_TO_VOID_FUNCTION =
       new Function<WriteLogEntriesResponse, Void>() {
         @Override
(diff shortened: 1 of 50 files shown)


google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
```diff
@@ -149,7 +154,7 @@ private static <V> V get(ApiFuture<V> future) {
       return Uninterruptibles.getUninterruptibly(future);
     } catch (ExecutionException ex) {
       Throwables.throwIfUnchecked(ex.getCause());
-      throw new RuntimeException(ex);
+      throw new VerifyException(ex);
```
losalex, line 157: I believe it might impact the existing contract with our library users - if user expecting `RuntimeException`, it will break

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
```diff
@@ -889,7 +896,7 @@ public void flush() {
     try {
       ApiFutures.allAsList(writesToFlush).get(FLUSH_WAIT_TIMEOUT_SECONDS, TimeUnit.SECONDS);
     } catch (InterruptedException | ExecutionException | TimeoutException e) {
-      throw new RuntimeException(e);
+      throw new VerifyException(e);
```
losalex, line 899: Same here - contact change for expected exception

losalex (commented):
Left a comment regarding changed exception thrown by code.

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
bpcreech: `VerifyException` is a child class of `RuntimeException`. If they have a `catch RuntimeException` this will still work.

I can imagine a circumstance in which they're checking for the exact class using getClass but I think that would be a really weird thing to do.

losalex: OK, I believe we can take this risk then

losalex (approved):

bpcreech:
Oh this was angering clirr because (per internal linter), I made a couple trivial breaking changes to the `Instrumentation` class. This class was holding only static things, so I marked it as `final` and made the ctor private to make it clear it shouldn't be instantiated. This is technically a breaking change! I could undo it if you want! Instead I marked it as an ignored difference for clirr because there should be no motivation for anyone to have ever instantiated that class to begin with.

losalex:
> Oh this was angering clirr because (per internal linter), I made a couple trivial breaking changes to the `Instrumentation` class. This class was holding only static things, so I marked it as `final` and made the ctor private to make it clear it shouldn't be instantiated. This is technically a breaking change! I could undo it if you want! Instead I marked it as an ignored difference for clirr because there should be no motivation for anyone to have ever instantiated that class to begin with.

Instrumentation is intended for internal usage only and making this final is OK if it doesn't breaks anything.

PR #1057, opened 2022-09-07
feat: Apply Google style recommendations

bpcreech:
This is a speculative PR just to show losalex@, will add issue if it seems like a good idea.

Fixes #TODO

Things our formatter does to this codebase:

1. Add `{ }` consistently after `if` statements. No one-liners-without-curlies allowed.
2. Group overloaded methods together
3. lowerCamelCase variables (`e_name` -> `eName`)
4. UPPER_SNAKE_CASE enums ~`Label.ClusterName` -> `Label.CLUSTER_NAME`). This would be a breaking change if the enum were intended to be public. I think `Label`'s not supposed to be (altho it is `protected` so it *could* be...). **I had to change the clirr allowlist for this!**. `Resource` is definitely private tho. **Update:** reverted the `Label` change because it's potentially breaking, but kept the `Resource` change.
5. Test methods in the form `lowerSnakeCaseOfSystemUnderTest_lowerSnakeCaseOfExpectedBehavior`
6. Don't do static import on classes (`import static com.google.cloud.logging.Logging.TailOption`).
7. Each top-level class gets its own file (`ResourceTypeEnvironmentGetterImpl`)

### DIFF
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
@@ -66,12 +66,12 @@ public class MonitoredResourceUtil {
   }
 
   private enum Resource {
-    CloudRun("cloud_run_revision"),
-    CloudFunction("cloud_function"),
-    AppEngine("gae_app"),
-    GceInstance("gce_instance"),
-    K8sContainer("k8s_container"),
-    Global("global");
+    CLOUD_RUN("cloud_run_revision"),
+    CLOUD_FUNCTION("cloud_function"),
+    APP_ENGINE("gae_app"),
+    GCE_INSTANCE("gce_instance"),
+    K8S_CONTAINER("k8s_container"),
+    GLOBAL("global");
 
     private final String key;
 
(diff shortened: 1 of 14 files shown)


google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -32,27 +32,27 @@
...
-    ModuleId("module_id"),
-    NamespaceName("namespace_name"),
-    PodName("pod_name"),
-    ProjectId("project_id"),
-    Region("region"),
-    RevisionName("revision_name"),
-    ServiceName("service_name"),
-    VersionId("version_id"),
-    Zone("zone");
+    CLUSTER_NAME("cluster_name"),
```
losalex, line 38: Indeed, the `MonitoredResourceUtil` is public and can be inherited - this means that `Label` could be accesses by child... I would be cautious here and perhaps leave this change out until we decide with major version update - let me know WDYT

losalex (commented):
In general, this looks good - thanks for your effort @bpcreech!
I just had one concern about changing labels - please take a look and let me know WDYT

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
bpcreech: yeah sounds good. i've reverted that part and removed the associated clirr exception

losalex (approved):

PR #1130, opened 2022-10-11
deps: regenerating with new Protobuf (3.7.x)

mpeddada1:
To merge into 3.7.x branch. This PR was generated using Protobuf 3.19.6. Source-Link: https://github.com/googleapis/googleapis/commit/c7ca416c2856aad6a3f7092924e56b8cf0cb2534

### DIFF
--- a/.kokoro/environment_tests.sh
+++ b/.kokoro/environment_tests.sh
@@ -27,6 +27,7 @@ fi
 # make sure submodule is up to date
 cd "$PROJECT_ROOT"
 git submodule update --init --recursive
+git config --global --add safe.directory /tmpfs/src/github/java-logging
 cd "${PROJECT_ROOT}/env-tests-logging"
 
 # Disable buffering, so that the logs stream through.
--- a/.kokoro/environment_tests.sh
+++ b/.kokoro/environment_tests.sh
@@ -27,6 +27,7 @@ fi
 # make sure submodule is up to date
 cd "$PROJECT_ROOT"
 git submodule update --init --recursive
+git config --global --add safe.directory '*'
 git config --global --add safe.directory /tmpfs/src/github/java-logging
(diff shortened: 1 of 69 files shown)


suztomo (approved):

losalex (commented):
Seems build is failed...

mpeddada1:
Hm adding `git config --global --add safe.directory '*'` and `git config --global --add safe.directory /tmpfs/src/github/java-logging` don't seem to do the trick.

.kokoro/environment_tests.sh
```diff
@@ -27,6 +27,7 @@ fi
 # make sure submodule is up to date
 cd "$PROJECT_ROOT"
 git submodule update --init --recursive
+git config --global --add safe.directory '*'
```
daniel-sanche, line 30: It looks like this should be moved up a line. It needs to run before `git submodule update`

mpeddada1: oh! good catch. I'll modify this.

daniel-sanche: It looks like this fixed that issue, but there's one more change you'll need to pull in from the `main` branch. It looks like the `python3.6` references in this file need to be changed to `python3` (3.6 is too out of date for the test container)

image

daniel-sanche: I'm not sure about the failing `GraalVM` tests though

losalex: thanks @daniel-sanche for comment - indeed, this seems as an issue!
Asd for `GraalVM`, I believe for now those can be ignored

mpeddada1: Applying this change seemed to have worked!

mpeddada1:
Thank you both so much for troubleshooting this! Merging this PR now.

Issue #1172, opened 2022-10-31
Make partialSuccess to be true by default


losalex:
Today `partialSuccess` does not set by default on [write](https://cloud.google.com/logging/docs/reference/v2/rest/v2/entries/write#body.request_body.FIELDS.partial_success) request which should be fixed in order to propagate as much logs as possible and only drop oversized entries

PR #1173, opened 2022-11-01
fix: Make partialSuccess to be true by default

losalex:
Fixes #[1172](https://github.com/googleapis/java-logging/issues/1172) ☕️

### DIFF
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
@@ -91,6 +91,7 @@ import com.google.protobuf.Empty;
 import com.google.protobuf.util.Durations;
 import java.text.ParseException;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -874,13 +875,14 @@ class LoggingImpl extends BaseService<LoggingOptions> implements Logging {
         logEntries =
             populateMetadata(logEntries, sharedResourceMetadata, this.getClass().getName());
       }
-      // Add partialSuccess = true option always for request which does not have it set explicitly.
-      // For request containing instrumentation data always override it with true.
-      writeLogEntries(
-          logEntries,
-          pair.x() || writeOptionPartialSuccessFlag == null
-              ? Instrumentation.addPartialSuccessOption(options)
-              : options);
+      // Add partialSuccess = true option always for request which does not have
+      // it set explicitly in options.
+      // For request containing instrumentation data (e.g. when pair.x() == true),
+      // always set or override partialSuccess with true.
+      if (pair.x() || writeOptionPartialSuccessFlag == null) {
+        options = addPartialSuccessOption(options);
+      }
+      writeLogEntries(logEntries, options);
       if (flushSeverity != null) {
         for (LogEntry logEntry : logEntries) {
           // flush pending writes if log severity at or above flush severity
@@ -1134,5 +1136,25 @@ class LoggingImpl extends BaseService<LoggingOptions> implements Logging {
   @VisibleForTesting
   int getNumPendingWrites() {
     return pendingWrites.size();
+  }
+
+  /**
+   * Adds a partialSuccess flag option to array of WriteOption
+   *
+   * @param options {WriteOption[]} The options array to be extended
+   * @return The new array of oprions containing WriteOption.OptionType.PARTIAL_SUCCESS flag set to
+   *     true
+   */
+  static WriteOption @Nullable [] addPartialSuccessOption(WriteOption[] options) {
+    if (options == null) {
+      return options;
+    }
+    List<WriteOption> writeOptions = new ArrayList<>();
+    Collections.addAll(writeOptions, options);
+    // Make sure we remove all partial success flags if any exist
+    writeOptions.removeIf(
+        option -> option.getOptionType() == WriteOption.OptionType.PARTIAL_SUCCESS);
+    writeOptions.add(WriteOption.partialSuccess(true));
+    return Iterables.toArray(writeOptions, WriteOption.class);
   }
 }
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
@@ -859,6 +859,8 @@ class LoggingImpl extends BaseService<LoggingOptions> implements Logging {
     try {
       final Map<Option.OptionType, ?> writeOptions = optionMap(options);
       final Boolean loggingOptionsPopulateFlag = getOptions().getAutoPopulateMetadata();
+      final Boolean writeOptionPartialSuccessFlag =
+          WriteOption.OptionType.PARTIAL_SUCCESS.get(writeOptions);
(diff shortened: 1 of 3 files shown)


google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
```diff
@@ -1869,6 +1870,7 @@ public void testWriteLogEntriesDoesNotEnableFlushByDefault() {
                     ImmutableList.of(
                         LOG_ENTRY1, LOG_ENTRY2.toBuilder().setSeverity(Severity.EMERGENCY).build()),
                     LogEntry.toPbFunction(PROJECT)))
+            .setPartialSuccess(true)
```
daniel-sanche, line 1873: Are there any tests in place that set it false?

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
```diff
@@ -872,9 +874,13 @@ public void write(Iterable<LogEntry> logEntries, WriteOption... options) {
         logEntries =
             populateMetadata(logEntries, sharedResourceMetadata, this.getClass().getName());
       }
-      // Add partialSuccess option always for request containing instrumentation data
+      // Add partialSuccess = true option always for request which does not have it set explicitly.
+      // For request containing instrumentation data always override it with true.
       writeLogEntries(
-          logEntries, pair.x() ? Instrumentation.addPartialSuccessOption(options) : options);
+          logEntries,
+          pair.x() || writeOptionPartialSuccessFlag == null
```
daniel-sanche, line 881: Isn't this missing the case where `writeOptionPartialSuccessFlag == true`?

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
```diff
@@ -2304,6 +2311,22 @@ public void testDiagnosticInfoWithPartialSuccess() {
...
+                Iterables.transform(
+                    ImmutableList.of(LOG_ENTRY1, LOG_ENTRY2), LogEntry.toPbFunction(PROJECT)))
+            .setPartialSuccess(false)
+            .build();
+    WriteLogEntriesResponse response = WriteLogEntriesResponse.getDefaultInstance();
+    EasyMock.expect(loggingRpcMock.write(request)).andReturn(ApiFutures.immediateFuture(response));
+    EasyMock.replay(rpcFactoryMock, loggingRpcMock);
+    logging = options.getService();
+    logging.write(ImmutableList.of(LOG_ENTRY1, LOG_ENTRY2), WriteOption.partialSuccess(false));
+  }
```
daniel-sanche, line 2328: should there be a duplicate test with `setPartialSuccess(true)`?

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
losalex: The idea is that if `writeOptionPartialSuccessFlag` present, do not override it since it was explicitly set by user

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
losalex: There are bunch of other tests in this file doing that

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
daniel-sanche: I read it a bit more. Is this equivalent?:

```
add_instrumentation = pair.x()
if (add_instrumentation || writeOptionPartialSuccessFlag == null || writeOptionPartialSuccessFlag == true ) {
    // Add partialSuccess = true option always for request which does not have it set explicitly, or
    // For any request containing instrumentation data
  options = Instrumentation.addPartialSuccessOption(options)
}
writeLogEntries(logEntries, options)
```

If so, I think something like this would make it more readable/maintainable.

I'd suggest possibly moving `addPartialSuccessOption` outside of `Instrumentation` too, since it's now being used by default

losalex: This logic is designed to **add** partialSuccess when it is not present in `options` and **override** `partialSuccess` with `true` for instrumentation record. I can move `Instrumentation.addPartialSuccessOption(options)` out to common

daniel-sanche: Hmm ok, I see what you're saying. This logic was very hard for me to parse though, would you be open to breaking up the `writeLogEntries` statement into multiple lines? There's a lot happening on one line here that can be hard to keep track of, especially using the `pair.x()` instead of a named variable

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingImplTest.java
daniel-sanche: Are you sure? The only other test I see that calls `logging.write` with a partial success option explicitly set is `testDiagnosticInfoGeneration`, but that's testing the case with a diagnostic entry, which follows a different flow through the code. 

It may be overkill, but it would be good to cover all the cases with the tests to be safe. Since the logic is a bit tricky here

losalex: New functionality is tested when `partialSuccess` it is not overriden if present in `options` - I believe this case is exactly tested here and it is enough since we do not care about a value, but rather that it is present or not.

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
losalex: Let me rewrite for clarity

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
```diff
@@ -872,9 +874,14 @@ public void write(Iterable<LogEntry> logEntries, WriteOption... options) {
...
             populateMetadata(logEntries, sharedResourceMetadata, this.getClass().getName());
       }
-      // Add partialSuccess option always for request containing instrumentation data
-      writeLogEntries(
-          logEntries, pair.x() ? Instrumentation.addPartialSuccessOption(options) : options);
+      // Add partialSuccess = true option always for request which does not have
+      // it set explicitly in options.
+      // For request containing instrumentation data (e.g. when pair.x() == true),
+      // always set or override partialSuccess with true.
+      if (pair.x() || writeOptionPartialSuccessFlag == null) {
```
daniel-sanche, line 881: nit: I think it would be more readable if you put `shouldAddInstrumentation = pair.x();` right under `logEntries = pair.y();`, so it's clear what this is

daniel-sanche (approved):

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
losalex: I believe the comment explains the meaning of `pair.x()`, so we can be good here without adding extra parameter.

PR #1179, opened 2022-11-03
feat: Add support for instrumentation version annotations

losalex:
Fixes #[1178](https://github.com/googleapis/java-logging/issues/1178) 🦕

### DIFF
--- a/.github/release-please.yml
+++ b/.github/release-please.yml
@@ -10,3 +10,5 @@ branches:
     handleGHRelease: true
     releaseType: java-backport
     branch: 3.7.x
+extraFiles:
+  - src/main/java/com/google/cloud/logging/Instrumentation.java
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/Instrumentation.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/Instrumentation.java
@@ -37,7 +37,7 @@ public final class Instrumentation {
   public static final String INSTRUMENTATION_NAME_KEY = "name";
   public static final String INSTRUMENTATION_VERSION_KEY = "version";
   public static final String JAVA_LIBRARY_NAME_PREFIX = "java";
-  public static final String DEFAULT_INSTRUMENTATION_VERSION = "UNKNOWN";
+  public static final String DEFAULT_INSTRUMENTATION_VERSION = "1.0.0"; // {x-release-please-version}
   public static final String INSTRUMENTATION_LOG_NAME = "diagnostic-log";
   public static final int MAX_DIAGNOSTIC_VALUE_LENGTH = 14;
   public static final int MAX_DIAGNOSTIC_ENTIES = 3;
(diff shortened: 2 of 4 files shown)


.github/release-please.yml
```diff
@@ -10,3 +10,5 @@ branches:
     handleGHRelease: true
     releaseType: java-backport
     branch: 3.7.x
+extraFiles:
+  - src/main/java/com/google/cloud/logging/Instrumentation.java
```
chingor13, line 14: this is per-configuration, so you'll want to add this to the other branch configurations (unless the file does not exist there).

google-cloud-logging/src/main/java/com/google/cloud/logging/Instrumentation.java
```diff
@@ -37,7 +37,8 @@ public final class Instrumentation {
   public static final String INSTRUMENTATION_NAME_KEY = "name";
   public static final String INSTRUMENTATION_VERSION_KEY = "version";
   public static final String JAVA_LIBRARY_NAME_PREFIX = "java";
-  public static final String DEFAULT_INSTRUMENTATION_VERSION = "UNKNOWN";
+  public static final String DEFAULT_INSTRUMENTATION_VERSION =
+      "1.0.0"; // {x-release-please-version}
```
chingor13, line 41: You probably want to use the java-specific replacement annotation: https://github.com/googleapis/java-logging/blob/b9e5824db9a294a898b0e84f26212300430c6526/google-cloud-logging/pom.xml#L6

chingor13 (changes requested):

.github/release-please.yml
losalex: Oh, you mean for LTS branch? If for LTS - we dont have to do it there since instrumentation is internal feature and I am OK with previous logic which supposed to work there

google-cloud-logging/src/main/java/com/google/cloud/logging/Instrumentation.java
losalex: Added an update - can you please let me know if thats the right way?

chingor13 (approved):

google-cloud-logging/src/main/java/com/google/cloud/logging/Instrumentation.java
```diff
@@ -37,7 +37,9 @@ public final class Instrumentation {
   public static final String INSTRUMENTATION_NAME_KEY = "name";
   public static final String INSTRUMENTATION_VERSION_KEY = "version";
   public static final String JAVA_LIBRARY_NAME_PREFIX = "java";
-  public static final String DEFAULT_INSTRUMENTATION_VERSION = "UNKNOWN";
+  // {x-version-update-start:google-cloud-logging:current}
+  public static final String DEFAULT_INSTRUMENTATION_VERSION = "1.0.0";
```
daniel-sanche, line 41: Are you sure 1.0.0 is a good default? Would 0.0.0 work?

daniel-sanche (approved):

google-cloud-logging/src/main/java/com/google/cloud/logging/Instrumentation.java
losalex: it supposed to change after release

PR #1196, opened 2022-11-15
fix: Wrong order of libraries info in instrumentation

losalex:
The instrumentation of each library should be appended to the end of list, so the inheriting library should be always first


daniel-sanche (approved):
might be good to add a test for this

Issue #1143, opened 2022-10-16
Flushing logs slows down the application intensively (Error severity logs are flushed by default) we need a way to disable flushing


montss:
#### Environment details

1. Java version: 11
2.  version(s): com.google.cloud:libraries-bom:26.1.3 --> com.google.cloud:google-cloud-logging-bom:3.11.5 --> com.google.cloud:google-cloud-logging:3.11.5

#### Steps to reproduce

  Using JUL

#### Code example

```java
		LoggingHandler googleHandler = new LoggingHandler();
		googleHandler.setFlushLevel(LoggingLevel.EMERGENCY);
		Logger logger = Logger.getLogger(TestClass.class.getName());
		logger.addHandler(googleHandler);
```

#### Any additional information below
Related to:-
https://github.com/googleapis/google-cloud-java/issues/2796
https://github.com/googleapis/google-cloud-java/pull/4254

Setting the flush level to something higher than error doesn't work.
Also setting it though logging.properties file doesn't work 
`com.google.cloud.logging.LoggingHandler.flushLevel=EMERGENCY`

Error logs are flushed by default which causes massive slows down for that request overall
Profiling the system shows that most of the time are gone as a wall time on the flush call

We need a way to disable flushing for good.
Didn't try .setFlushLevel(Level.OFF);

losalex:
Thanks @montss for opening this issue. Just to clarify - calling `setFlushLevel(Level.EMERGENCY)` in code was not working for you, correct?

montss:
> Thanks @montss for opening this issue. Just to clarify - calling `setFlushLevel(Level.EMERGENCY)` in code was not working for you, correct?

Correct, calling `setFlushLevel(Level.EMERGENCY)` didn't work nor setting it through the `logging.properties` file 
`com.google.cloud.logging.LoggingHandler.flushLevel=EMERGENCY`

daniel-sanche:
I don't have much context on this, but possibly related issue: https://github.com/googleapis/java-logging-logback/pull/319

losalex:
> I don't have much context on this, but possibly related issue: [googleapis/java-logging-logback#319](https://github.com/googleapis/java-logging-logback/pull/319)

Thanks @daniel-sanche for context! The one issue you shared indeed related to java-logback and is not relevant for java-logging.

losalex:
@montss , I checked in a fix - would you please be able to test it and let me know it works for you?

losalex:
[...]

PR #1206, opened 2022-11-23
fix: Need a way to disable flushing

losalex:
Fixes #[1143](https://github.com/googleapis/java-logging/issues/1143) ☕️

### DIFF
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
@@ -498,31 +498,21 @@ public class LoggingHandler extends Handler {
     if (level instanceof LoggingLevel) {
       return ((LoggingLevel) level).getSeverity();
     }
-
-    switch (level.intValue()) {
-        // FINEST
-        // FINER
-        // FINE
-      case 300:
-      case 400:
-      case 500:
-        return Severity.DEBUG;
-        // CONFIG
-        // INFO
-      case 700:
-      case 800:
-        return Severity.INFO;
-        // WARNING
-      case 900:
-        return Severity.WARNING;
-        // SEVERE
-      case 1000:
-        return Severity.ERROR;
-      case Integer.MAX_VALUE:
-        return Severity.UNRECOGNIZED;
-      default:
-        return Severity.DEFAULT;
+    // Choose the severity based on Level range.
+    // The assumption is that Level values below maintain same numeric value
+    int value = level.intValue();
+    if (value <= Level.FINE.intValue()) {
+      return Severity.DEBUG;
+    } else if (value <= Level.INFO.intValue()) {
+      return Severity.INFO;
+    } else if (value <= Level.WARNING.intValue()) {
+      return Severity.WARNING;
+    } else if (value <= Level.SEVERE.intValue()) {
+      return Severity.ERROR;
+    } else if (value == Level.OFF.intValue()) {
+      return Severity.NONE;
     }
+    return Severity.DEFAULT;
   }
 
   private static boolean isTrueOrNull(Boolean b) {
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
@@ -518,6 +518,8 @@ public class LoggingHandler extends Handler {
         // SEVERE
       case 1000:
         return Severity.ERROR;
+      case Integer.MAX_VALUE:
+        return Severity.UNRECOGNIZED;
       default:
         return Severity.DEFAULT;
     }
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
@@ -882,7 +882,7 @@ class LoggingImpl extends BaseService<LoggingOptions> implements Logging {
         options = Instrumentation.addPartialSuccessOption(options);
       }
       writeLogEntries(logEntries, options);
-      if (flushSeverity != Severity.NONE) {
+      if (flushSeverity != null && flushSeverity != Severity.NONE) {
         for (LogEntry logEntry : logEntries) {
           // flush pending writes if log severity at or above flush severity
           if (logEntry.getSeverity().compareTo(flushSeverity) >= 0) {
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
@@ -882,7 +882,7 @@ class LoggingImpl extends BaseService<LoggingOptions> implements Logging {
         options = Instrumentation.addPartialSuccessOption(options);
       }
       writeLogEntries(logEntries, options);
-      if (flushSeverity != null) {
+      if (flushSeverity != null && flushSeverity != Severity.UNRECOGNIZED) {
         for (LogEntry logEntry : logEntries) {
           // flush pending writes if log severity at or above flush severity
           if (logEntry.getSeverity().compareTo(flushSeverity) >= 0) {
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
@@ -882,6 +882,8 @@ class LoggingImpl extends BaseService<LoggingOptions> implements Logging {
         options = Instrumentation.addPartialSuccessOption(options);
       }
       writeLogEntries(logEntries, options);
+      // We treat Severity.UNRECOGNIZED as an option to disable flushing completely,
+      // since LogSeverity does not exposed OFF as a severity level
       if (flushSeverity != null && flushSeverity != Severity.UNRECOGNIZED) {
         for (LogEntry logEntry : logEntries) {
           // flush pending writes if log severity at or above flush severity
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
@@ -107,7 +107,7 @@ class LoggingImpl extends BaseService<LoggingOptions> implements Logging {
   private final Map<Object, ApiFuture<Void>> pendingWrites = new ConcurrentHashMap<>();
 
   private volatile Synchronicity writeSynchronicity = Synchronicity.ASYNC;
-  private volatile Severity flushSeverity = null;
+  private volatile Severity flushSeverity = Severity.NONE;
   private boolean closed;
 
   private static Boolean emptyToBooleanFunction(Empty input) {
@@ -882,9 +882,7 @@ class LoggingImpl extends BaseService<LoggingOptions> implements Logging {
         options = Instrumentation.addPartialSuccessOption(options);
       }
       writeLogEntries(logEntries, options);
-      // We treat Severity.UNRECOGNIZED as an option to disable flushing completely,
-      // since LogSeverity does not exposed OFF as a severity level
-      if (flushSeverity != null && flushSeverity != Severity.UNRECOGNIZED) {
+      if (flushSeverity != Severity.NONE) {
         for (LogEntry logEntry : logEntries) {
           // flush pending writes if log severity at or above flush severity
           if (logEntry.getSeverity().compareTo(flushSeverity) >= 0) {
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
@@ -138,7 +138,12 @@ class LoggingImpl extends BaseService<LoggingOptions> implements Logging {
 
   @Override
   public void setFlushSeverity(Severity flushSeverity) {
-    this.flushSeverity = flushSeverity;
+    // For backward compatibility we treat null as Severity.NONE
+    if (flushSeverity == null) {
+      this.flushSeverity = Severity.NONE;
+    } else {
+      this.flushSeverity = flushSeverity;
+    }
   }
 
   @Override
@@ -882,7 +887,7 @@ class LoggingImpl extends BaseService<LoggingOptions> implements Logging {
         options = Instrumentation.addPartialSuccessOption(options);
       }
       writeLogEntries(logEntries, options);
-      if (flushSeverity != null && flushSeverity != Severity.NONE) {
+      if (flushSeverity != Severity.NONE) {
         for (LogEntry logEntry : logEntries) {
           // flush pending writes if log severity at or above flush severity
           if (logEntry.getSeverity().compareTo(flushSeverity) >= 0) {
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/Severity.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/Severity.java
@@ -26,8 +26,8 @@ import com.google.logging.type.LogSeverity;
  */
 public enum Severity {
 
-  /** The unrecognized severity level. Should not be used with log entries. */
-  UNRECOGNIZED(LogSeverity.UNRECOGNIZED),
+  /** The None severity level. Should not be used with log entries. */
+  NONE(LogSeverity.UNRECOGNIZED),
 
   /** The log entry has no assigned severity level. */
   DEFAULT(LogSeverity.DEFAULT),
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/Severity.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/Severity.java
@@ -26,7 +26,7 @@ import com.google.logging.type.LogSeverity;
  */
 public enum Severity {
 
-  /** The log entry has unrecognized severity level. */
+  /** The unrecognized severity level. Should not be used with log entries. */
   UNRECOGNIZED(LogSeverity.UNRECOGNIZED),
 
   /** The log entry has no assigned severity level. */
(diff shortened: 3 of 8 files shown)


daniel-sanche:
Can you provide some more context to make review easier? It looks like you're trying to provide a way to avoid flushing logs. How does Severity.UNRECOGNIZED tie into this?

Note that this sounds very similar to a [bug I addressed in the logback library](https://github.com/googleapis/java-logging-logback/pull/319) a couple years ago. Does this tie in with that feature in any way?

losalex:
> Can you provide some more context to make review easier? It looks like you're trying to provide a way to avoid flushing logs. How does Severity.UNRECOGNIZED tie into this?
> 
> Note that this sounds very similar to a [bug I addressed in the logback library](https://github.com/googleapis/java-logging-logback/pull/319) a couple years ago. Does this tie in with that feature in any way?

I use `Severity.UNRECOGNIZED` as a severity signaling to turn off flushing for `java-logging` since there is no way to extend equivalent `LogSeverity` which is `protobuf` enum . The fix you shared is related to `java-logback` and not related to `java-logging` which can be used standalone. Other than that the change is straightforward - exposing setters/getters and patching logic to avoid flushing

daniel-sanche:
> I use `Severity.UNRECOGNIZED` as a severity to signal turning off flushing for `java-logging` since there is no way to extend equivalent `LogSeverity` which is `protobuf` enum .

Overloading `Severity.UNRECOGNIZED` for this doesn't really sit right with me. It feels like it could cause significant technical debt in the future. 
1. Can we use a separate bool flag to turn flushing on/off instead of overloading the Severity field?
2. Or, since everything seems to use the Severity enum, couldn't we add a Severity.OFF item there, without it having a corresponding proto entry?

losalex:
> > I use `Severity.UNRECOGNIZED` as a severity to signal turning off flushing for `java-logging` since there is no way to extend equivalent `LogSeverity` which is `protobuf` enum .
> 
> Overloading `Severity.UNRECOGNIZED` for this doesn't really sit right with me. It feels like it could cause significant technical debt in the future.
> 
> 1. Can we use a separate bool flag to turn flushing on/off instead of overloading the Severity field?
> 2. Or, since everything seems to use the Severity enum, couldn't we add a Severity.OFF item there, without it having a corresponding proto entry?

I must admit I am not exactly following your suggestions above - the feature for flush severity is designed to flush pending writes if log severity at or above flush severity...  `LogSeverity.UNRECOGNIZED` already exists and I use it only for this specific feature to indicate that we should skip logging since flush severity is `Severity.UNRECOGNIZED` - agree that it might not be great name, but I was thinking that creating `Severity.OFF = LogSeverity.UNRECOGNIZED` did not make much sense either since `Severity` is tied with `LogSeverity`.

daniel-sanche:
> pending writes if log severity is equal or above flush severity... `LogSeverity.UNRECOGNIZED` already exists and I use it only for this specific feature to indicate that we should skip logging since flush severity

My concern is that the spec for LogSeverity.UNRECOGNIZED was not intended to mean "log flusing off", so I'm worried about future problems from overloading the meaning of that field

> but I was thinking that creating `Severity.OFF = LogSeverity.UNRECOGNIZED` did not make much sense either since `Severity` is tied with `LogSeverity`.

Could we do `Severity.OFF = null` instead, so we're not overloading an existing field?

losalex:
> > pending writes if log severity is equal or above flush severity... `LogSeverity.UNRECOGNIZED` already exists and I use it only for this specific feature to indicate that we should skip logging since flush severity
> 
> My concern is that the spec for LogSeverity.UNRECOGNIZED was not intended to mean "log flusing off", so I'm worried about future problems from overloading the meaning of that field
> 
> > but I was thinking that creating `Severity.OFF = LogSeverity.UNRECOGNIZED` did not make much sense either since `Severity` is tied with `LogSeverity`.
> 
> Could we do `Severity.OFF = null` instead, so we're not overloading an existing field?

I guess none of `Severity` fields were intended to be used in o


> > pending writes if log severity is equal or above flush severity... `LogSeverity.UNRECOGNIZED` already exists and I use it only for this specific feature to indicate that we should skip logging since flush severity
> 
> My concern is that the spec for LogSeverity.UNRECOGNIZED was not intended to mean "log flusing off", so I'm worried about future problems from overloading the meaning of that field
> 
> > but I was thinking that creating `Severity.OFF = LogSeverity.UNRECOGNIZED` did not make much sense either since `Severity` is tied with `LogSeverity`.
> 
> Could we do `Severity.OFF = null` instead, so we're not overloading an existing field?

>> My concern is that the spec for LogSeverity.UNRECOGNIZED was not intended to mean "log flusing off", so I'm worried about future problems from overloading the meaning of that field

Not sure I understand what you mean here - can you please let me know what is a meaning of `LogSeverity.UNRECOGNIZED`?

daniel-sanche:
Unrecognized is part of the [log severity proto definition](https://googleapis.github.io/common-protos-java/1.14.0/apidocs/com/google/logging/type/LogSeverity.html). The spec doesn't make it clear how it is meant to be used, but that doesn't mean it's safe to redefine. 

We will likely encounter some legitimate logs with their severity field set to "unrecognized". (speculation: maybe some OTel exporters will use "unrecognized" when there's not a direct mapping between a GCP LogSeverity and a different vendor's?) If we treat "Unrecognized" to mean "Log flushing Off", that could make things confusing if customers attempt to actually filter for those logs. And it could be hard to fix in the future

google-cloud-logging/src/main/java/com/google/cloud/logging/Severity.java
```diff
@@ -26,6 +26,9 @@
  */
 public enum Severity {
 
+  /** The unrecognized severity level. Should not be used with log entries. */
+  UNRECOGNIZED(LogSeverity.UNRECOGNIZED),
```
minherz, line 30: there is already an existing severity for the kind of "undefined/unrecognized" type of severity which is `Severity.DEFAULT`.

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -518,6 +518,8 @@ private static Severity severityFor(Level level) {
         // SEVERE
       case 1000:
         return Severity.ERROR;
+      case Integer.MAX_VALUE:
+        return Severity.UNRECOGNIZED;
```
minherz, line 522: nit: this method's original implementation is bad because it tries to use the internal knowledge about `class java.util.logging.Level` constants to make the conversion when it should actually substitute the Level's constants with the Severity's constants.
The change introduces an additional complexity by enforcing the "closed" range for "default" values which previously was bound only from the bottom (even despite the fact that the top boundary is set to the integer's maximal value).

losalex:
> 

> Unrecognized is part of the [log severity proto definition](https://googleapis.github.io/common-protos-java/1.14.0/apidocs/com/google/logging/type/LogSeverity.html). The spec doesn't make it clear how it is meant to be used, but that doesn't mean it's safe to redefine.
> 
> We will likely encounter some legitimate logs with their severity field set to "unrecognized". (speculation: maybe some OTel exporters will use "unrecognized" when there's not a direct mapping between a GCP LogSeverity and a different vendor's?) If we treat "Unrecognized" to mean "Log flushing Off", that could make things confusing if customers attempt to actually filter for those logs. And it could be hard to fix in the future

>> In order to redefine a meaning, we need to know the meaning... `LogSeverity.UNRECOGNIZED` used in library already today to indicate invalid severity levels and end up in exception (see [here](https://github.com/googleapis/java-logging/blob/79e9d8d00b0b9be10a12d6bb37f387f6f8b356f2/google-cloud-logging/src/main/java/com/google/cloud/logging/Severity.java#L86) for example). Also it cannot be used to generate log entries due to same reason - severity `UNRECOGNIZED` is not supported. So I am less worried about exposing `Severity.UNRECOGNIZED` since we can clearly document that it cannot be used to generate log entries and if anyone uses it the exception will happen. And if my assumption is correct, seems `Severity.UNRECOGNIZED` will never be used to generate legit logs since it is common practice to define values in enum for invalid/error cases.
In general, we do not have an equivalent for `Level.OFF` in `Severity` and using `LogSeverity.UNRECOGNIZED` is probably the only option since Java does not allows enum inheritance...

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
```diff
@@ -882,7 +882,9 @@ public void write(Iterable<LogEntry> logEntries, WriteOption... options) {
         options = Instrumentation.addPartialSuccessOption(options);
       }
       writeLogEntries(logEntries, options);
-      if (flushSeverity != null) {
+      // We treat Severity.UNRECOGNIZED as an option to disable flushing completely,
+      // since LogSeverity does not exposed OFF as a severity level
+      if (flushSeverity != null && flushSeverity != Severity.UNRECOGNIZED) {
```
minherz, line 887: The implementation tries to use `enum Severity` for more than one goal: to define the severity level **and** to define on/off conditions for flushing. I do not think it is a good idea.

minherz (changes requested):
Given the expectations to the backward compatibility and an attempt to avoid introduction of the new interfaces while keeping the existing interface intuitive, I propose the following implementation and recommend to extend it to the java-logging-logback library:

1. Refactor `LoggingHandler.severityFor(Level level)` method to switch/case based on the `java.util.logging.Level` values and not `Integer` values that supposed to reflect the numeric values of the logging levels.
2. For the `case Level.OFF` return null:
   ```java
   case Level.OFF:
       return null;
   ```
3. Keep the rest implementation (I mainly reference to [LoggingImpl.java:885][1]) as it is now.

[1]: https://github.com/googleapis/java-logging/blob/34b20a277d22174907329a900918e9a24dbdd8de/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java#L885

google-cloud-logging/src/main/java/com/google/cloud/logging/Severity.java
losalex: > Given the expectations to the backward compatibility and an attempt to avoid introduction of the new interfaces while keeping the existing interface intuitive, I propose the following implementation and recommend to extend it to the java-logging-logback library:
> 
> 1. Refactor `LoggingHandler.severityFor(Level level)` method to switch/case based on the `java.util.logging.Level` values and not `Integer` values that supposed to reflect the numeric values of the logging levels.
> 2. For the `case Level.OFF` return null:
>    ```java
>    case Level.OFF:
>        return null;
>    ```
> 3. Keep the rest implementation (I mainly reference to [LoggingImpl.java:885](https://github.com/googleapis/java-logging/blob/34b20a277d22174907329a900918e9a24dbdd8de/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java#L885)) as it is now.

This is actually going to break backward compatibility since returning `null` from this method never happened before...

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
losalex: I guess `Level` does the same - it combines `OFF` with other levels...

google-cloud-logging/src/main/java/com/google/cloud/logging/Severity.java
losalex: > there is already an existing severity for the kind of "undefined/unrecognized" type of severity which is `Severity.DEFAULT`.

`Severity.DEFAULT` is recognized today by Logging as appropriate level and considered as default, while `LogSeverity.UNRECOGNIZED` is not.

losalex: It seems that `Level` is a class and I am planning to refactor `LoggingHandler.severityFor(Level level)` to work for ranges

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
losalex: Changed this to ranges evaluation

minherz (approved):
very solid improvement compared to the previous push.
consider adding the mentioned `null` validations.
do you think adding a test that validates the exception if trying to write with `Severity.NONE` makes sense here?

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -498,29 +498,21 @@ private static Severity severityFor(Level level) {
     if (level instanceof LoggingLevel) {
       return ((LoggingLevel) level).getSeverity();
     }
```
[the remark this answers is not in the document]
losalex: Nope, it doesn't

google-cloud-logging/src/main/java/com/google/cloud/logging/Severity.java
```diff
@@ -26,6 +26,9 @@
  */
 public enum Severity {
 
+  /** The None severity level. Should not be used with log entries. */
+  NONE(LogSeverity.UNRECOGNIZED),
+
```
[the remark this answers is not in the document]
losalex: Unfortunately, the `Severity` constructor accepts only `LogSeverity` values... Not sure how else to solve rather than adding another constructor - let me add it then.

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -498,29 +498,21 @@ private static Severity severityFor(Level level) {
...
-      case 900:
-        return Severity.WARNING;
-        // SEVERE
-      case 1000:
-        return Severity.ERROR;
-      default:
-        return Severity.DEFAULT;
+    // Choose the severity based on Level range.
+    // The assumption is that Level values below maintain same numeric value
+    int value = level.intValue();
```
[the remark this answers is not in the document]
losalex: Thanks :-)
As for `level == null` - I can definitely add it, but it will make this code to be null tolerant which is in turn could have an impact on other functions like `logEntryFor()`. I guess I will leave it as is

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
```diff
@@ -882,7 +882,7 @@ public void write(Iterable<LogEntry> logEntries, WriteOption... options) {
         options = Instrumentation.addPartialSuccessOption(options);
       }
       writeLogEntries(logEntries, options);
-      if (flushSeverity != null) {
+      if (flushSeverity != Severity.NONE) {
```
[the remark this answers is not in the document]
losalex: yep, will do

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingHandlerTest.java
```diff
@@ -537,6 +543,37 @@ public void testFlushLevel() {
     handler.publish(newLogRecord(Level.WARNING, MESSAGE));
   }
 
+  @Test
+  public void testFlushLevelOff() {
+    logging.setFlushSeverity(Severity.NONE);
+    expectLastCall().anyTimes();
```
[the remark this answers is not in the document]
losalex: Oh, sure

losalex:
> very solid improvement compared to the previous push. consider adding the mentioned `null` validations. do you think adding a test that validates the exception if trying to write with `Severity.NONE` makes sense here?

Thanks! Added test to validate that `Severity.NONE` cannot be used

losalex:
> > Given the expectations to the backward compatibility and an attempt to avoid introduction of the new interfaces while keeping the existing interface intuitive, I propose the following implementation and recommend to extend it to the java-logging-logback library:
> > 
> > 1. Refactor `LoggingHandler.severityFor(Level level)` method to switch/case based on the `java.util.logging.Level` values and not `Integer` values that supposed to reflect the numeric values of the logging levels.
> > 2. For the `case Level.OFF` return null:
> >    ```java
> >    case Level.OFF:
> >        return null;
> >    ```
> > 3. Keep the rest implementation (I mainly reference to [LoggingImpl.java:885](https://github.com/googleapis/java-logging/blob/34b20a277d22174907329a900918e9a24dbdd8de/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java#L885)) as it is now.
> 
> Yeah, this seems like a great solution to me. It seems to be in line with the [original intent of the setFlushSeverity](https://github.com/googleapis/java-logging/blob/34b20a277d22174907329a900918e9a24dbdd8de/google-cloud-logging/src/main/java/com/google/cloud/logging/Logging.java#L305) function, where null was documented to mean "flush off", so this feels like the safest way to implement this change

As I mentioned before, I believe that using nullable enums is not a good practice. Added `Severity.NONE`

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -498,29 +498,21 @@ private static Severity severityFor(Level level) {
...
+    } else if (value <= Level.INFO.intValue()) {
+      return Severity.INFO;
+    } else if (value <= Level.WARNING.intValue()) {
+      return Severity.WARNING;
+    } else if (value <= Level.SEVERE.intValue()) {
+      return Severity.ERROR;
+    } else if (value == Level.OFF.intValue()) {
+      return Severity.NONE;
     }
+    return Severity.DEFAULT;
```
daniel-sanche, line 515: Probably not related to this change, but there are three severities above error: [Critical, Alert, and  Emergency](https://googleapis.github.io/common-protos-java/1.14.0/apidocs/com/google/logging/type/LogSeverity.html). Would it make sense to add them here in some way in the future?

google-cloud-logging/src/main/java/com/google/cloud/logging/Severity.java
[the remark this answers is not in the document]
daniel-sanche: I'm still a bit concerned about using UNRECOGNIZED for this, in case we need to include UNRECOGNIZED as a true option for this enum in the future. But assuming that UNRECOGNIZED exists only in the java protos and not in the proto definitions, it should be ok

daniel-sanche (approved):

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
losalex: Yes, we can - depend on ask of course. I do not have any strong opinion here :-)

Issue #987, opened 2022-07-04
Java-Logging API: Retrieving logentries pagewise always results in an exception


mwolff-bay:
OS-Type: MacOS Darwin 21.5.0
Java-Version: openjdk version "1.8.0_322"

Hi folks,
please find attached a small demo for creating and retrieving logentries (a readme is included).
Whenever I try to iterate over the first page of retrieved logentries I get the following exception:

Exception in thread "main" com.google.cloud.logging.LoggingException: io.grpc.StatusRuntimeException: INVALID_ARGUMENT: page_token doesn't match arguments from the request
	at com.google.cloud.logging.spi.v2.GrpcLoggingRpc$2.apply(GrpcLoggingRpc.java:208)
	at com.google.cloud.logging.spi.v2.GrpcLoggingRpc$2.apply(GrpcLoggingRpc.java:202)
	at com.google.api.core.ApiFutures$ApiFunctionToGuavaFunction.apply(ApiFutures.java:239)
	at com.google.common.util.concurrent.AbstractCatchingFuture$CatchingFuture.doFallback(AbstractCatchingFuture.java:234)
	at com.google.common.util.concurrent.AbstractCatchingFuture$CatchingFuture.doFallback(AbstractCatchingFuture.java:222)
	at com.google.common.util.concurrent.AbstractCatchingFuture.run(AbstractCatchingFuture.java:133)
	at com.google.common.util.concurrent.DirectExecutor.execute(DirectExecutor.java:31)
	at com.google.common.util.concurrent.AbstractFuture.executeListener(AbstractFuture.java:1270)
	at com.google.common.util.concurrent.AbstractFuture.complete(AbstractFuture.java:1038)
	at com.google.common.util.concurrent.AbstractFuture.setException(AbstractFuture.java:808)
	at com.google.api.gax.retrying.BasicRetryingFuture.handleAttempt(BasicRetryingFuture.java:200)
	at com.google.api.gax.retrying.CallbackChainRetryingFuture$AttemptCompletionListener.handle(CallbackChainRetryingFuture.java:135)
	at com.google.api.gax.retrying.CallbackChainRetryingFuture$AttemptCompletionListener.run(CallbackChainRetryingFuture.java:117)
	at com.google.common.util.concurrent.DirectExecutor.execute(DirectExecutor.java:31)
	at com.google.common.util.concurrent.AbstractFuture.executeListener(AbstractFuture.java:1270)
[...]

PR #1220, opened 2022-12-06
fix: Retrieving logentries pagewise always results in an exception

losalex:
Fixes #[987](https://github.com/googleapis/java-logging/issues/987) ☕️

### DIFF
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
@@ -1126,17 +1126,17 @@ class LoggingImpl extends BaseService<LoggingOptions> implements Logging {
   }
 
   static Map<Option.OptionType, ?> updateFilter(final Map<Option.OptionType, ?> options) {
+    // We should see if filter provided in otiopns have a timestamp parameter
+    // and if not, it should be added with further update of options map.
     String existingFilter = EntryListOption.OptionType.FILTER.get(options);
     String newFilter = generateFilter(existingFilter);
     if (newFilter.equals(existingFilter)) {
       return options;
     }
-    Map<Option.OptionType, Object> optionMap = Maps.newHashMap();
-    for (Map.Entry<Option.OptionType, ?> entry : options.entrySet()) {
-      optionMap.put(entry.getKey(), entry.getValue());
-    }
-    optionMap.put(EntryListOption.OptionType.FILTER, newFilter);
-    return optionMap;
+    // Update
+    Map<Option.OptionType, Object> optionsCopy = Maps.newHashMap(options);
+    optionsCopy.put(EntryListOption.OptionType.FILTER, newFilter);
+    return optionsCopy;
   }
 
   static String generateFilter(String filter) {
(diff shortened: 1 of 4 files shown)


google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
```diff
@@ -1137,6 +1125,37 @@ public void close() throws Exception {
     return optionMap;
   }
 
+  static Map<Option.OptionType, ?> updateFilter(final Map<Option.OptionType, ?> options) {
+    String existingFilter = EntryListOption.OptionType.FILTER.get(options);
```
daniel-sanche, line 1129: nit: some function comments would be helpful

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
```diff
@@ -1137,6 +1125,37 @@ public void close() throws Exception {
...
+    String newFilter = generateFilter(existingFilter);
+    if (newFilter.equals(existingFilter)) {
+      return options;
+    }
+    Map<Option.OptionType, Object> optionMap = Maps.newHashMap();
+    for (Map.Entry<Option.OptionType, ?> entry : options.entrySet()) {
+      optionMap.put(entry.getKey(), entry.getValue());
+    }
```
daniel-sanche, line 1137: nit: is there a reason to avoid using [this constructor](https://guava.dev/releases/19.0/api/docs/com/google/common/collect/Maps.html#newHashMap(java.util.Map))?
 `Map<Option.OptionType, Object> optionMap = Maps.newHashMap(options);`

Also the variable name optionMap vs options is a bit confusing

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingTest.java
```diff
@@ -100,6 +104,18 @@ public void testEntryListOption() {
...
+    Map<Option.OptionType, ?> options = LoggingImpl.optionMap(EntryListOption.filter(FILTER));
+    assertThat((String) EntryListOption.OptionType.FILTER.get(options)).isEqualTo(FILTER);
+    Map<Option.OptionType, ?> updated = LoggingImpl.updateFilter(options);
+    assertTrue(((String) EntryListOption.OptionType.FILTER.get(updated)).contains("timestamp"));
+    assertFalse(options == updated);
+    assertNotEquals(EntryListOption.OptionType.FILTER.get(updated), FILTER);
+    Map<Option.OptionType, ?> anotherUpdated = LoggingImpl.updateFilter(updated);
+    assertTrue(anotherUpdated == updated);
+  }
+
```
daniel-sanche, line 118: Would it be helpful to add a test that does paging like the linked error report?

daniel-sanche (approved):

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
losalex: Oh, great comment - thank you! Changed the code to use the proper constructor and also variable name.

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
losalex: Added

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingTest.java
losalex: Most of our tests are done with mocks and apparently there is no much point of simulating such error (e.g. this code does not handle the error but rather tries to prevent the problem). I believe this would be better covered by environment tests to make sure paging works as expected

PR #1364, opened 2023-06-06
fix: Remove org.jspecify dependency

### LINKED ISSUE #1362: fix: Remove org.jspecify dependency
Based on https://github.com/jspecify/jspecify#readme, the usage of the org.jspecify annotations are not recommended in release libraries, so it needs to be removed from code for now.

losalex:
Fixes [#1362](https://github.com/googleapis/java-logging/issues/1362) ☕️

### DIFF
--- a/google-cloud-logging/pom.xml
+++ b/google-cloud-logging/pom.xml
@@ -91,11 +91,6 @@
       <artifactId>google-cloud-core</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.jspecify</groupId>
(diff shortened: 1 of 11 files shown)


suztomo:
Would you replace the occurrences with `javax.annotation.Nullable`?

losalex:
> Would you replace the occurrences with `javax.annotation.Nullable`?

Discussed offline - for now removing those annotations until we have a stable version of **org.jspecify**

suztomo (approved):
The javax.annotation.Nullable won't work for the internal environment.

PR #1507, opened 2024-01-11
chore: remove obsolete release job configs

alicejli:
Release config is now managed internally.

The requirements.txt and requirements.in files are now unused here.

Note that I had to add entries in owlbot.py because they are templated files.

Related to b/296905003

Based on https://github.com/googleapis/java-firestore/pull/1404


suztomo (approved):

suztomo:
requirements.txt is not part of this change?

alicejli:
> requirements.txt is not part of this change?

Good catch - have updated.

PR #1521, opened 2024-01-25
ci: unmanaged dependency check

suztomo:
Cindy has reviewed go/cloud-sdk-java-dependency-governance-design. This new check would fail if someone adds an unknown new dependency to pom.xml files. All new dependencies should be added in https://github.com/googleapis/sdk-platform-java/blob/main/java-shared-dependencies/third-party-dependencies/pom.xml first.

### DIFF
--- a/.github/workflows/unmanaged_dependency_check.yaml
+++ b/.github/workflows/unmanaged_dependency_check.yaml
@@ -1,0 +1,22 @@
+on:
+  pull_request:
+name: Unmanaged dependency check
+jobs:
+  unmanaged_dependency_check:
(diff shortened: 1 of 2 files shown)


suztomo:
Unmanaged dependency check / unmanaged_dependency_check (pull_request) Successful

suztomo:
@cindy-peng This is the check discussed in go/cloud-sdk-java-dependency-governance-design and we are gradually rolling out to the repositories.

cindy-peng (approved):
Thanks Tomo for working on this!

PR #1661, opened 2024-07-26
[test] protobuf to 3.25.4 for v3.17.1

zhumin8:
test for lts7


zhumin8:
graalvm test failure due to quota. will retry.

PR #1729, opened 2024-11-19
feat: introduce `java.time` methods

diegomarquezp:
This PR introduces `java.time` alternatives to existing `org.threeten.bp.*` methods, as well as switching internal variables (if any) to `java.time`

The main constraint is to keep the changes backwards compatible, so for each existing threeten method "`method1(org.threeten.bp.Duration)`" we will add an alternative with a _Duration_ (or _Timestamp_ when applicable) suffix: "`method1Duration(java.time.Duration)`".

For most cases, the implementation will be held in the `java.time` method and the old threeten method will just delegate the call to it. However, for the case of abstract classes, the implementation will be kept in the threeten method to avoid breaking changes (i.e. users that already overloaded the method in their user code).

### DIFF
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
@@ -34,6 +34,7 @@ import com.google.logging.v2.LogEntryOperation;
 import com.google.logging.v2.LogEntrySourceLocation;
 import com.google.logging.v2.LogName;
 import java.io.Serializable;
+import java.time.Duration;
 import java.time.Instant;
 import java.util.HashMap;
 import java.util.Map;
@@ -612,6 +613,14 @@ public class LogEntry implements Serializable {
     }
   }
 
+  static final class DurationSerializer implements JsonSerializer<Duration> {
+    @Override
+    public JsonElement serialize(
+        Duration src, java.lang.reflect.Type typeOfSrc, JsonSerializationContext context) {
+      return new JsonPrimitive(src.toString());
+    }
+  }
+
   static final class SourceLocationSerializer implements JsonSerializer<SourceLocation> {
     @Override
     public JsonElement serialize(
@@ -649,6 +658,7 @@ public class LogEntry implements Serializable {
       checkNotNull(builder);
       this.gson =
           new GsonBuilder()
+              .registerTypeAdapter(Duration.class, new DurationSerializer())
               .registerTypeAdapter(Instant.class, new InstantSerializer())
               .registerTypeAdapter(SourceLocation.class, new SourceLocationSerializer())
               .registerTypeAdapter(HttpRequest.RequestMethod.class, new RequestMethodSerializer())
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
@@ -34,7 +34,6 @@ import com.google.logging.v2.LogEntryOperation;
 import com.google.logging.v2.LogEntrySourceLocation;
 import com.google.logging.v2.LogName;
 import java.io.Serializable;
-import java.time.Duration;
 import java.time.Instant;
 import java.util.HashMap;
 import java.util.Map;
@@ -613,14 +612,6 @@ public class LogEntry implements Serializable {
     }
   }
 
-  static final class DurationSerializer implements JsonSerializer<Duration> {
-    @Override
-    public JsonElement serialize(
-        Duration src, java.lang.reflect.Type typeOfSrc, JsonSerializationContext context) {
-      return new JsonPrimitive(src.toString());
-    }
-  }
-
   static final class SourceLocationSerializer implements JsonSerializer<SourceLocation> {
     @Override
     public JsonElement serialize(
@@ -658,7 +649,6 @@ public class LogEntry implements Serializable {
       checkNotNull(builder);
       this.gson =
           new GsonBuilder()
-              .registerTypeAdapter(Duration.class, new DurationSerializer())
               .registerTypeAdapter(Instant.class, new InstantSerializer())
               .registerTypeAdapter(SourceLocation.class, new SourceLocationSerializer())
               .registerTypeAdapter(HttpRequest.RequestMethod.class, new RequestMethodSerializer())
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/HttpRequest.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/HttpRequest.java
@@ -16,14 +16,18 @@
 
 package com.google.cloud.logging;
 
+import static com.google.api.gax.util.TimeConversionUtils.toJavaTimeDuration;
+import static com.google.api.gax.util.TimeConversionUtils.toThreetenDuration;
+
 import com.google.api.core.ApiFunction;
+import com.google.api.core.ObsoleteApi;
 import com.google.cloud.StringEnumType;
 import com.google.cloud.StringEnumValue;
 import com.google.common.base.MoreObjects;
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
 import java.io.Serializable;
+import java.time.Duration;
 import java.util.Objects;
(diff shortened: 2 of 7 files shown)


google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
```diff
@@ -612,6 +613,14 @@ public JsonElement serialize(
     }
   }
 
+  static final class DurationSerializer implements JsonSerializer<Duration> {
+    @Override
+    public JsonElement serialize(
+        Duration src, java.lang.reflect.Type typeOfSrc, JsonSerializationContext context) {
+      return new JsonPrimitive(src.toString());
+    }
+  }
```
lqiu96, line 622: qq, why is this DurationSerializer added?

diegomarquezp: By default, the Gson serializer will try to access private methods/fields of Duration (and Instant - see the already-existing seralizer above). This seralizer relies on `toString()` instead.

lqiu96: > By default, the Gson serializer will try to access private methods/fields of Duration

I'm not following this part. Why does it try to access these methods + fields from Duration? Also, why is this an issue now and not previously?

diegomarquezp:
Samples/checkstyle has been failing in other PRs as well

see https://github.com/googleapis/java-logging/actions/runs/11917969552/job/33214317931?pr=1729
see https://github.com/googleapis/java-logging/pull/1728

google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
diegomarquezp: > Why does it try to access these methods + fields from Duration? 
I think it's meant to create a JSON string by accessing all its fields (i.e. `{"seconds":123,"nanos":123}`

> Also, why is this an issue now and not previously?

I'm trying to figure this out. The JDK [version](https://github.com/adoptium/jdk8u/blob/ed3d731386bf4f6819d5e5f69376e520df3a6262/jdk/src/share/classes/java/time/Duration.java#L159-L164) has both fields as private. [So does](https://github.com/ThreeTen/threetenbp/blob/b833efe7ac2f1a02c016deb188fd7ce3b124ed2e/src/main/java/org/threeten/bp/Duration.java#L127-L132) the backport. Both also implement the same interfaces (`Serializable`, `Comparable`, `TemporalAmount`.
The failing call happens in GSON when it tries to make the fields accessible - see below.

> Is this related to an error message/ stacktrace from a test?
```
com.google.gson.JsonIOException: Failed making field 'java.time.Duration#seconds' accessible; either increase its visibility or write a custom TypeAdapter for its declaring type.
See https://github.com/google/gson/blob/main/Troubleshooting.md#reflection-inaccessible

	at com.google.gson.internal.reflect.ReflectionHelper.makeAccessible(ReflectionHelper.java:76)
	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory.getBoundFields(ReflectiveTypeAdapterFactory.java:388)
	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory.create(ReflectiveTypeAdapterFactory.java:161)
	at com.google.gson.Gson.getAdapter(Gson.java:628)
	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory.createBoundField(ReflectiveTypeAdapterFactory.java:201)
	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory.getBoundFields(ReflectiveTypeAdapterFactory.java:395)
	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory.create(ReflectiveTypeAdapterFactory.java:161)
	at com.google.gson.Gson.getAdapter(Gson.java:628)
	at com.google.gson.Gson.toJson(Gson.java:928)
	at com.google.gson.Gson.toJson(Gson.java:899)
	at com.google.gson.Gson.toJson(Gson.java:848)
	at com.google.gson.Gson.toJson(Gson.java:825)
	at com.google.cloud.logging.LogEntry$StructuredLogFormatter.appendField(LogEntry.java:683)
	at com.google.cloud.logging.LogEntry$StructuredLogFormatter.appendField(LogEntry.java:693)
	at com.google.cloud.logging.LogEntry.toStructuredJsonString(LogEntry.java:737)
	at com.google.cloud.logging.LogEntryTest.testStructureLogPresentations(LogEntryTest.java:385)



Caused by: java.lang.reflect.InaccessibleObjectException: Unable to make field private final long java.time.Duration.seconds accessible: module java.base does not "opens java.time" to unnamed module @6996db8
	at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:354)
	at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:297)
	at java.base/java.lang.reflect.Field.checkCanSetAccessible(Field.java:178)
	at java.base/java.lang.reflect.Field.setAccessible(Field.java:172)
	at com.google.gson.internal.reflect.ReflectionHelper.makeAccessible(ReflectionHelper.java:68)
	... 42 more
```
I'm not sure what this means: `java.base does not "opens java.time"`

diegomarquezp: I believe it has something to do with modules? If this behaves perfectly fine if I run the test on java 8.

diegomarquezp: According to [this SO](https://stackoverflow.com/questions/70376339/gson-unable-to-serialize-exception-with-java-17), it's by design: https://bugs.openjdk.org/browse/JDK-8256358

lqiu96: Ok gotcha. Thanks for take a look into this.

Can you see if doing `--add-opens java.base/java.time=ALL-UNNAMED` would work? Something similar to https://github.com/googleapis/java-datastore/blob/8244ad258a90e5bbc7e7d8ec384555b2be84ab8f/pom.xml#L251. If this does work, I think that could be an option to consider. 

Also, can you add a link for that to the PR Description as an explanation for why we are adding this. 

The rest LGTM and I think we just need Cindy to review as well.

lqiu96 (approved):
LGTM. Cindy to review for final approval. Adding a Do not merge label until then.

google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
diegomarquezp: Ok, add-opens works but in a different way:
 * In maven-jar-plugin, we declare a manifest entry that will be detected in Java 9+ _only_. This is basically the same `--add-opens` flag and is packed inside the jar as MANIFEST.mf
 * Since maven-surefire-plugin does not use the jar directly, I also added a surefire config to include this flag when running the tests

diegomarquezp: I oversaw this brings trouble in units(8). Looking into it

diegomarquezp: Looks like we are having a limitation. The following code activates the `--add-opens` flag in tests when the _jdk_ version is >= 9:
https://github.com/googleapis/java-logging/blob/5a5abd436102c1fc8d89ef09ea829739cea310d5/pom.xml#L305-L317

However, when running the CIs, we specify the runtime via the `-Djvm=...` option (absolute path to runtime), which varies among environments - see
https://github.com/googleapis/java-logging/blob/5a5abd436102c1fc8d89ef09ea829739cea310d5/.github/workflows/ci.yaml#L46-L58

and

https://github.com/googleapis/java-logging/blob/5a5abd436102c1fc8d89ef09ea829739cea310d5/.kokoro/build.sh#L49-L53

IIUC we would need a regex matcher for the `<activation>` entry to only include "paths not containing '1.8'".

diegomarquezp: I'll be reverting this maven config in the meantime

cindy-peng (approved):

